PR #21, opened 2018-08-17
Duplicate Jenkinsfiles for Uyuni until we can use parameters

juliogonzalez:


jenkins_pipelines/uyuni_prs/tests_files/rubocop.sh
```diff
@@ -0,0 +1,5 @@
+#! /bin/bash
```
MalloZup: @juliogonzalez  if we go opensource and the repo is public, we might run this via travis, i think we could consider it.
in

juliogonzalez: IMHO we should use the same stuff for both Uyuni and SUSE manager, to avoid confusing people.

MalloZup: ok

juliogonzalez:
Thanks! Merging now.

PR #22, opened 2018-09-12
Enable PR merging on changelog tests

juliogonzalez:
So after the changelog test passes (it doesn't matter the result), merging for the PR will be enabled.

### DIFF
--- a/jenkins_pipelines/manager_prs/Jenkinsfile_prs_changelog
+++ b/jenkins_pipelines/manager_prs/Jenkinsfile_prs_changelog
@@ -10,19 +10,24 @@ pipeline {
 
   // env variables
   environment { 
-    gitarro_changelog_check = "gitarro.ruby2.1  -r SUSE/spacewalk" + 
+    repository = "SUSE/spacewalk"
+    gitarro_changelog_check = "gitarro.ruby2.1  -r ${repository}" + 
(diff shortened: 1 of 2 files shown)


jenkins_pipelines/manager_prs/Jenkinsfile_prs_changelog
```diff
@@ -41,7 +46,11 @@ pipeline {
         stage('Run changelo test') {
             steps {
                 echo 'Run changelog tests'
-                sh "${gitarro_changelog_test}"
+                sh "${gitarro_changelog_test};" +
+                   "CLRET=\$?;" +
+                   "PR_NUMBER=\$(grep 'GITARRO_PR_NUMBER:' ${env.WORKSPACE}/.gitarro_vars | cut -d ':' -f2);" +
+                   "if [ \$CLRET -eq 0 ]; then ${gitarro_enable_merging} -t /usr/bin/true -P \$PR_NUMBER; else ${gitarro_enable_merging} -t /usr/bin/false -P \$PR_NUMBER; fi;" +
```
juliogonzalez: Just a not for the future: cloning the repo again just to run /usr/bin/true is a waste of time and resources.

It would be good if we could tell gitarro to avoid cloning, or if we could switch two statuses with one single test :-)

PR #27, opened 2018-11-20
Hotfix/6001 cleanup always

srbarrios:
Task : https://github.com/SUSE/spacewalk/issues/6001
- PR Pipelines must clean workspace when success, and keep it when fails.
- Always clean workspace when starts


MalloZup:
I would not use a library it will make things more complicated, speaking from, caasp experience

MalloZup:
A library in groovy is not beautiful to maintain and it will be pain for people who wants to create a simple pipeline

MalloZup:
Have look here https://github.com/kubic-project/automation/tree/master/jenkins-pipelines imho a library is just overkill. The logic should go to the scripts the scheduler call not in jenkins imh

PR #30, opened 2018-11-22
[6376] Checkout ci repo after cleanws, to fix rubocop not found

srbarrios:
Task: https://github.com/SUSE/spacewalk/issues/6376

### Fixing rubocop job
This job needs files inside susemanager-ci repo, but the step cleanws at first stage is deleting also the repo checked out. There is no way to clean up before the checkout of pipelines repo, we don't want to use deletedir of a specific folder because it can be tricky, as pipelines also creates tmp and script dirs, only cleaned with cleanws step.

Solution:
- Checkout ci repo after cleanws, to fix rubocop script not found.
- Also added a bonus, boolean parameter in case you want to run the pipeline manually, avoiding clean up workspace after job success.

### DIFF
--- a/jenkins_pipelines/manager_prs/Jenkinsfile_prs_backend_unitpsql
+++ b/jenkins_pipelines/manager_prs/Jenkinsfile_prs_backend_unitpsql
@@ -28,15 +28,22 @@ pipeline {
    }
    // run only on specific hosts
     agent { label 'suse-manager-unit-tests' }
-   // trigger
+    
+    parameters {
+        booleanParam(defaultValue: true, description: 'Clean up workspace after a successful execution.', name: 'cleanWorkspace')
+    }
+    
     triggers {
         cron('H/5 * * * *')
-       }
+    }
+    
     stages {
         stage('Clean Up Workspace') {
             steps {
                 echo 'Clean up previous workspace'
                 cleanWs()
+                echo 'Check out SCM'
+                checkout scm
             }
         }
         stage('Check Pull Request') {
@@ -58,9 +65,11 @@ pipeline {
         }
     }
   post { 
-        success {
-            echo 'Clean up current workspace, when job success.'
-            cleanWs()
+        script {
+            if (params.cleanWorkspace == true) {
+                echo 'Clean up current workspace, when job success.'
+                cleanWs()
+            }
         }
   }
 }
--- a/jenkins_pipelines/manager_prs/Jenkinsfile_prs_changelog
+++ b/jenkins_pipelines/manager_prs/Jenkinsfile_prs_changelog
@@ -30,7 +30,11 @@ pipeline {
    }
    // run only on specific hosts
     agent { label 'suse-manager-unit-tests' }
-   // trigger
+    
+    parameters {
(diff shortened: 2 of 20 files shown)


jenkins_pipelines/manager_prs/Jenkinsfile_prs_backend_unitpsql
```diff
@@ -28,15 +28,22 @@ pipeline {
...
         cron('H/5 * * * *')
-       }
+    }
+    
     stages {
         stage('Clean Up Workspace') {
             steps {
                 echo 'Clean up previous workspace'
                 cleanWs()
+                echo 'Check out SCM'
```
juliogonzalez: FMPOV, checking out should be a second step.

jenkins_pipelines/manager_prs/Jenkinsfile_prs_backend_unitpsql
```diff
@@ -58,9 +65,11 @@ pipeline {
         }
     }
   post { 
-        success {
```
juliogonzalez: Isn't this going to run even when the test fails?

It's not clear to me by having a look at https://jenkins.io/doc/book/pipeline/syntax/#post

IMHO a better approach (and easier to understand) would be:

```
post { 
      success {
            if (params.cleanWorkspace == true) {
                echo 'Clean up current workspace, when job succeeds.'
                cleanWs()
            }
      }
}
```

juliogonzalez:
Eeeeeeeeh.... I had comments to do! :-(

srbarrios:
Ups! Though is was ok! Replying now :)

jenkins_pipelines/manager_prs/Jenkinsfile_prs_backend_unitpsql
srbarrios: So, the checkout is done at the beginning of the job. But as we do a cleanWs, this checkout is deleted and to fix it, we checkout again here. But it gonna be ugly in the pipeline to have two times the checkout stage, it's way is a second step in the same stage.

juliogonzalez: Ah, right, damned pipelines...

jenkins_pipelines/manager_prs/Jenkinsfile_prs_backend_unitpsql
srbarrios: You are right, it should be only when success, that's a mistake.
Anyhow the correct way is:

````
  post { 
        success{
            script {
                if (params.cleanWorkspace == true) {
                    echo 'Clean up current workspace, when job success.'
                    cleanWs()
                }
            }
        }
  }
```

juliogonzalez: No problem :-)

I didn't made corrections there because I suddenly saw the new PR :-\

MalloZup:
good cath @juliogonzalez , @srbarrios  no pb, as thumb rule wait maximal a day for People to review . ( unless it is weekend :robot: )

MalloZup:
or christmas :christmas_tree:

PR #33, opened 2019-01-02
Feature/6411 unified pooling pipeline

srbarrios:
Task : https://github.com/SUSE/spacewalk/issues/6411

It creates a new pipeline that will do pooling every 5 min. checkinf for pending PRs, if a PR is pending it will trigger all the checks (the other pipelines).
It's made to reduce the amount of request we currently do to github pooling for changes, but most important it will give us a cleaner history in the jenkins jobs of PR checks, as they will be only ran when there are actually something to do.


srbarrios:
Ready, please take a look to the new pipeline.
You have also the job working here: https://ci.suse.de/view/Manager/view/Manager-prs/job/manager-prs-trigger/

PR #35, opened 2019-02-21
Remove gitarro checks and allow test gitarro from a PR

srbarrios:
Do you have an example working here:
https://ci.suse.de/view/Manager/view/Manager-prs/job/manager-prs-check-develop/

Tha changes are only made to main pipeline, I will need to see how to do the same for the rest of pipelines propagating the parameter PR_NUMBER


jenkins_pipelines/manager_prs/Jenkinsfile_prs_backend_unittests_pgsql
```diff
@@ -2,72 +2,79 @@
...
-                  sh "${runtest}"
-                  junit "**/backend/reports/pgsql_tests.xml" 
+                echo 'Run tests!'
+                script {
+                    commands = "${gitarro_cmd} ${runtest}"
+                    if (params.GITARRO_PR_NUMBER != "") {
+                            commands = "${gitarro_local} ${runtest}"
+                    }
+                }
+                sh "set +e; ${commands}; TESTS_RESULT=\$?; set -e;"
```
juliogonzalez: Why do we need this?

jenkins_pipelines/manager_prs/Jenkinsfile_prs_backend_unittests_pgsql
```diff
@@ -2,72 +2,79 @@
...
+                echo 'Collecting JUnit Test reports'
+                junit allowEmptyResults: true, testResults: "**/backend/reports/pgsql_tests.xml"
```
juliogonzalez: Ah, ok... I guess because you want to collect junit tests even if there were failures.

Makes sense,but next time please do separate PRs :-)

srbarrios: No, so the point here is that previously, we checked if the test must be passed with a "gitarro check" execution, now we rid of that part. As that check is done in main pipeline, but adding this parameter "allowEmptyResults" we prevent the build from failing in case the test is skipped.

srbarrios:
> Looks good.
> 
> Just two things:
> 
> 1. Next time please consider separate PRs for fixing the format, and for bugfixes/features unrelated to the PR (so it's easier to review).
> 2. I really think we should, in the future, refactor and create a library or something that the jobs can use, to avoid having duplicated code making the code easier to maintain.

+1000

PR #42, opened 2019-02-26
Improve Jenkins Pipelines to reduce GitHub calls

srbarrios:
- Added a PR_NUMBER parameter in PR Check jobs, to be able to run on a specific PR directly
- Add the PR_NUMBER to the title of PR Check jobs executed
- Fix rubocop pipeline test that needs a script contained in pipelines repo

### DIFF
--- a/jenkins_pipelines/manager_prs/Jenkinsfile_prs_check
+++ b/jenkins_pipelines/manager_prs/Jenkinsfile_prs_check
@@ -9,7 +9,7 @@ properties([
 pipeline {
 
     parameters {
-        string(defaultValue: "", description: 'Gitarro PR', name: 'GITARRO_PR_NUMBER')
+        string(defaultValue: '', description: 'Gitarro PR', name: 'GITARRO_PR_NUMBER')
     }
 
     environment {
@@ -22,7 +22,7 @@ pipeline {
     agent { label 'suse-manager-unit-tests' }
 
     triggers {
-        cron('H/5 * * * *')
+        cron('H/10 * * * *')
     }
 
     stages {
@@ -33,7 +33,7 @@ pipeline {
                 echo 'Check out SCM'
                 checkout scm
                 script {
-                    if (params.GITARRO_PR_NUMBER != "") {
+                    if (params.GITARRO_PR_NUMBER != '') {
                         echo 'Check out Gitarro PR'
                         checkout([$class           : 'GitSCM', branches: [[name: "FETCH_HEAD"]],
                                   extensions       : [[$class: 'LocalBranch']],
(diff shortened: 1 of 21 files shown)


jenkins_pipelines/manager_prs/Jenkinsfile_prs_check
```diff
@@ -9,7 +9,7 @@ properties([
 pipeline {
 
     parameters {
-        string(defaultValue: "", description: 'Gitarro PR', name: 'GITARRO_PR_NUMBER')
+        string(defaultValue: '167', description: 'Gitarro PR', name: 'GITARRO_PR_NUMBER')
```
juliogonzalez: Why do we want `167` as default value?

jenkins_pipelines/manager_prs/Jenkinsfile_prs_ruby_rubocop
```diff
@@ -49,9 +52,13 @@ pipeline {
                 echo 'Run tests!'
                 script {
                     commands = "${gitarro_cmd} ${runtest}"
```
juliogonzalez: Not for this PR (in fact I should have commented this before), but why not having:
```
                    if (params.GITARRO_PR_NUMBER != '') {
                            commands = "${gitarro_local} ${runtest}"
                    }
                    else
                            commands = "${gitarro_cmd} ${runtest}"
                    }
```
?

jenkins_pipelines/uyuni_prs/Jenkinsfile_prs_check
```diff
@@ -9,7 +9,7 @@ properties([
 pipeline {
 
     parameters {
-        string(defaultValue: "", description: 'Gitarro PR', name: 'GITARRO_PR_NUMBER')
+        string(defaultValue: '167', description: 'Gitarro PR', name: 'GITARRO_PR_NUMBER')
```
juliogonzalez: Same as before.

jenkins_pipelines/manager_prs/Jenkinsfile_prs_check
srbarrios: We don't. That's what I told you we need to keep in mind and remove just after merge gitarro, so meanwhile is Jenkins CI is working fine using that fork and that gitarro changes

jenkins_pipelines/manager_prs/Jenkinsfile_prs_ruby_rubocop
srbarrios: Just because if we have that, we need to have a previous line declaring the variable with something. Otherwise, the pipeline will not be happy.

jenkins_pipelines/uyuni_prs/Jenkinsfile_prs_check
srbarrios: yep

srbarrios:
REMINDER: Before merge it, gitarro PR167 must be merged, gem installed and then remove the 167 number as default parameter en main pipeline.

PR #43, opened 2019-03-14
Increase the time to check a PR to 2 days old

srbarrios:
Increasing from 1h to 2days the filter of PRs to check.
As we sometimes have issues in our Jenkins slaves, and that cause that we need to force-push our PR to be checked.
We have around 5-6 PR max younger than 2 days, so it should be fine. 
And it will safe us from some support discussions on IRC.


srbarrios:
So, for that PR is just good with one person ;)

PR #45, opened 2019-05-13
New force_test parameter on merge_enabled tets, so it will run always

srbarrios:
New force_test parameter added on changelog_test pipelines for Uyuni and SUMA, so it will run always the merged_enabled test even if we don't have a comment or a checkbox to re-trigger it.
This is to fix a corner case, when the changelog_test fails, we mark the merged_enabled to false, then we re-run the changelog_test with the checkbox but the merged_enabled doesn't change its status (as it is not triggered because has not a checkbox marked in the PR description)


srbarrios:
We need first to merge and have a gem of https://github.com/openSUSE/gitarro/pull/169
Then we can merge that one.

srbarrios:
Gitarro merged. Let's merge it too.

PR #49, opened 2019-07-10
Finishing stage moved to always, junit, remove sumaform_env param

srbarrios:
On that PR we move some final actions from a regular stage to a post-stage, which is always triggered.

In addition:
-  we include junit test collection.
- we remove `sumaform_env` parameter, as having it, the pipeline will override the value set manually on each job. We will use the same pipeline for both NUE and PROVO env.

### DIFF
--- a/jenkins_pipelines/manager_prs/Jenkinsfile_testsuite
+++ b/jenkins_pipelines/manager_prs/Jenkinsfile_testsuite
@@ -6,11 +6,12 @@ properties([
     disableConcurrentBuilds(),
 ])
 
+def deployed = false
+
 pipeline {
 
     parameters {
         string(defaultValue: 'Manager-3.2', description: 'Testsuite GitHub branch: It should match the value at manager-3.2-2obs', name: 'testsuite_branch')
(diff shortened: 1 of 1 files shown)


jenkins_pipelines/manager_prs/Jenkinsfile_testsuite
```diff
@@ -6,11 +6,12 @@ properties([
     disableConcurrentBuilds(),
 ])
 
+def deployed = false
+
 pipeline {
 
     parameters {
         string(defaultValue: 'Manager-3.2', description: 'Testsuite GitHub branch: It should match the value at manager-3.2-2obs', name: 'testsuite_branch')
-        string(defaultValue: '32/VARS-full-NUE.sh', description: 'Sumaform Test Runner environment to use', name: 'sumaform_env')
```
juliogonzalez: Why?

srbarrios: > LGTM. Just one question... and something I just noticed: we have this at `manager_prs ` folder, but it doesn't make much sense (anyway fixing it is not for this PR).

At that moment it only runs on Manager-3.2, it supposed to be identical for Uyuni, parametrizing it properly. But let's keep it here, until we do the transition to Uyuni.

PR #51, opened 2019-07-31
Reorganize folders

srbarrios:
When we created the testsuite pipeline, we store it in manager_prs, which is wrong.
This PR creates new folders for manager and uyuni testsuites, where we can include our pipelines for QA/QAM testsuites.


srbarrios:
Not merge please, only approve :)
As I will need to change a jenkins job configuration at the same time it's merged.

PR #55, opened 2019-09-17
Fix the package used by javascript_lint from js/ to web/

lneves12:
Fix the package used by javascript_lint from js/ to web/


lneves12:
done :)

PR #57, opened 2019-10-08
WIP: Add script to create and revert snapshots

ktsamis:
Add script to create and revert snapshots
Fixes https://github.com/SUSE/spacewalk/issues/9544


jenkins_pipelines/manager_testsuite/Jenkinsfile_qa_testsuite
```diff
@@ -41,26 +41,32 @@ pipeline {
                     } else {
```
srbarrios: I would say we need to remove all the snapshots from all stages from previous Jenkins run before deploy.
Otherwise, it could fail during `terraform apply` or other not wanted scenarios.

jenkins_pipelines/manager_testsuite/Jenkinsfile_qa_testsuite
```diff
@@ -41,26 +41,32 @@ pipeline {
...
+                    sh "bash create-snapshots.sh create_snapshots deploy"
+                    deployed = true
                 }
             }
         }
 
         stage('Core - Setup') {
             steps {
                 script { env.deployed = true }
+                sh "bash create-snapshots.sh revert_snapshots deploy"
```
srbarrios: We need to are more logic on reverts, revert always is a waste of time, we need to revert only when : 
- The stage is re-started
- The snapshot exist

jenkins_pipelines/manager_testsuite/Jenkinsfile_qa_testsuite
ktsamis: Now added a cleanup at the end of the pipeline, don't think it's possible in the beginning of the pipeline to do cleanup of the previous.

jenkins_pipelines/manager_testsuite/Jenkinsfile_qa_testsuite
ktsamis: I added a counter and conditionals to check if it restarted. 
If the snapshot doesn't exist it means it failed to create it before at the stage and the pipeline stopped there. We won't be able to rerun the next stage at all if the previous failed

srbarrios:
@ktsamis please resolve the conflict, squash all your commits in one and test the branch in a job.
When the PR is in good shape, re-add me as reviewer :)

ktsamis:
This is dead

PR #60, opened 2019-11-06
Drop OpenStack based configurations

moio:
This is in preparation for https://github.com/uyuni-project/sumaform/pull/631

Rationale:
 - cucumber_per_prs/31: disabled since March 2018, never succeeded once
in history
 - cucumber_per_prs/head: disabled and put in Obsoleted since the same
date. Also never succeeded
 - performance_testsuite: abandoned since long time, not succeeding since
about 2 years


moio:
FTR decision has been taken and I have approval from our Director

ktsamis:
I think it's good to remove now. It creates more overhead to maintain and we can always decide to readd the functionality in the future depending on the needs of the team. Since it's not working anyway let's remove now.

PR #71, opened 2020-03-30
Move sanity check to a separate stage

srbarrios:
Card: SUSE/spacewalk#10347

We want to have the sanity checks in a different stage of the testsuite, so if the sanity checks fails, the rest of the test suite is aborted.

Related PRs:

- Manager-3.2 SUSE/spacewalk#11101
- Manager-4.0 SUSE/spacewalk#11099
- Uyuni https://github.com/uyuni-project/uyuni/pull/2073

### DIFF
--- a/jenkins_pipelines/manager_testsuite/Jenkinsfile_qa_sequential_testsuite
+++ b/jenkins_pipelines/manager_testsuite/Jenkinsfile_qa_sequential_testsuite
@@ -57,6 +57,12 @@ pipeline {
             }
         }
 
+        stage('Sanity Check') {
+            steps {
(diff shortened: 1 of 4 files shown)


srbarrios:
As commented in the chat, I think we should do it differently.
We are depending on the deployment stage, just because we made the sanity checks using our test framework.
Instead, IMHO, we should have an independent script, which runs before the deployment stage.

@Bischoff as you code the first implementation, what do you think about that refactor?

srbarrios:
NEVERMIND :trollface: 
Let's merge it what we planned. This refactor has very low priority and don't make even sense to waste time discussing it now.

Bischoff:
@srbarrios I don't have strong feelings about this refactor.

But as said in chat, this PR was probably missing a change in the yaml files.

PR #72, opened 2020-04-03
Jenkinsfile with terracumber

juliogonzalez:
Tested with 3.2 (full and refenv), Test-Naica and Uyuni-Master.

This change, besides using sumaform, allows pipelines with separate parameters for each environment.

The target is removing `manager_testsuite` and `uyuni_testsuite` after everything is migrated.

### DIFF
--- a/jenkins_pipelines/environments/common/pipeline-reference.groovy
+++ b/jenkins_pipelines/environments/common/pipeline-reference.groovy
@@ -1,0 +1,42 @@
+def run(params) {
+    timestamps {
+        deployed = false
+        env.resultdir = "${WORKSPACE}/results"
+        env.resultdirbuild = "${resultdir}/${BUILD_NUMBER}"
+        env.common_params = "--outputdir ${resultdir} --tf ${params.tf_file} --gitfolder ${resultdir}/sumaform"
+        try {
+            stage('Clone terracumber, susemanager-ci and sumaform') {
+                // Create a directory for  to place the directory with the build results (if it does not exist)
+                sh "mkdir -p ${resultdir}"
+                git url: params.terracumber_gitrepo, branch: params.terracumber_ref
+                dir("susemanager-ci") {
+                    checkout scm
+                }
+                // Clone sumaform
+                sh "set +x; source /home/jenkins/.credentials set -x; ./terracumber-cli ${common_params} --gitrepo ${params.sumaform_gitrepo} --gitref ${params.sumaform_ref} --runstep gitsync"
+            }
+            stage('Deploy') {
+                // Provision the environment
+                if (params.terraform_bin_init) {
+                    env.TERRAFORM_INIT = '--init'
+                } else {
+                    env.TERRAFORM_INIT = ''
+                }
+                sh "set +x; source /home/jenkins/.credentials set -x; TERRAFORM=${params.terraform_bin} TERRAFORM_PLUGINS=${params.terraform_bin_plugins} ./terracumber-cli ${common_params} --logfile ${resultdirbuild}/sumaform.log ${env.TERRAFORM_INIT} --taint '.*(domain|main_disk).*' --runstep provision"
+                deployed = true
+            }
+        }
+        finally {
+            stage('Get results') {
+                if (!deployed) {
+                  // Send email
+                  sh "./terracumber-cli ${common_params} --logfile ${resultdirbuild}/mail.log --runstep mail"
+                }
+                // Clean up old results
+                sh "./clean-old-results -r ${resultdir}"
+            }
+        }
+    }
+}
+
+return this
--- a/jenkins_pipelines/environments/common/pipeline.groovy
+++ b/jenkins_pipelines/environments/common/pipeline.groovy
@@ -1,0 +1,84 @@
+def run(params) {
+    timestamps {
+        deployed = false
+        env.resultdir = "${WORKSPACE}/results"
+        env.resultdirbuild = "${resultdir}/${BUILD_NUMBER}"
+        // The junit plugin doesn't affect full paths
+        junit_resultdir = "results/${BUILD_NUMBER}/results_junit"
+        env.common_params = "--outputdir ${resultdir} --tf ${params.tf_file} --gitfolder ${resultdir}/sumaform"
+        try {
+            stage('Clone terracumber, susemanager-ci and sumaform') {
+                // Create a directory for  to place the directory with the build results (if it does not exist)
+                sh "mkdir -p ${resultdir}"
+                git url: params.terracumber_gitrepo, branch: params.terracumber_ref
+                dir("susemanager-ci") {
+                    checkout scm
+                }
+                // Clone sumaform
+                sh "set +x; source /home/jenkins/.credentials set -x; ./terracumber-cli ${common_params} --gitrepo ${params.sumaform_gitrepo} --gitref ${params.sumaform_ref} --runstep gitsync"
+            }
+            stage('Deploy') {
+                // Provision the environment
+                if (params.terraform_bin_init) {
+                    env.TERRAFORM_INIT = '--init'
+                } else {
+                    env.TERRAFORM_INIT = ''
+                }
+                sh "set +x; source /home/jenkins/.credentials set -x; TERRAFORM=${params.terraform_bin} TERRAFORM_PLUGINS=${params.terraform_bin_plugins} ./terracumber-cli ${common_params} --logfile ${resultdirbuild}/sumaform.log ${env.TERRAFORM_INIT} --taint '.*(domain|main_disk).*' --runstep provision"
+                deployed = true
+            }
+            stage('Sanity Check') {
+                sh "./terracumber-cli ${common_params} --logfile ${resultdirbuild}/testsuite.log --runstep cucumber --cucumber-cmd 'cd /root/spacewalk/testsuite; rake cucumber:sanity_check'"
+            }
+            
+            stage('Core - Setup') {
+                sh "./terracumber-cli ${common_params} --logfile ${resultdirbuild}/testsuite.log --runstep cucumber --cucumber-cmd 'cd /root/spacewalk/testsuite; rake cucumber:core'"
+                sh "./terracumber-cli ${common_params} --logfile ${resultdirbuild}/testsuite.log --runstep cucumber --cucumber-cmd 'cd /root/spacewalk/testsuite; rake cucumber:reposync'"
+            }
+            stage('Core - Initialize clients') {
+                sh "./terracumber-cli ${common_params} --logfile ${resultdirbuild}/testsuite.log --runstep cucumber --cucumber-cmd 'cd /root/spacewalk/testsuite; rake cucumber:init_clients'"
+            }
+            stage('Secondary features') {
+                def statusCode1 = sh script:"./terracumber-cli ${common_params} --logfile ${resultdirbuild}/testsuite.log --runstep cucumber --cucumber-cmd 'cd /root/spacewalk/testsuite; rake cucumber:secondary'", returnStatus:true
+                def statusCode2 = sh script:"./terracumber-cli ${common_params} --logfile ${resultdirbuild}/testsuite.log --runstep cucumber --cucumber-cmd 'cd /root/spacewalk/testsuite; rake cucumber:secondary_parallelizable'", returnStatus:true
+                sh "exit \$(( ${statusCode1}|${statusCode2} ))"
+            }
+        }
+        finally {
+            stage('Get results') {
+                def error = 0
+                if (deployed) {
+                    try {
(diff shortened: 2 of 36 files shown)


jenkins_pipelines/environments/common/pipeline-reference.groovy
```diff
@@ -0,0 +1,44 @@
+def run(params) {
+    timestamps {
+        deployed = false
+        env.resultdir = "${WORKSPACE}/results"
+        env.resultdirbuild = "${resultdir}/${BUILD_NUMBER}"
+        env.common_params = "--outputdir ${resultdir} --tf ${params.tf_file} --gitfolder ${resultdir}/sumaform"
+        try {
+            stage('Clone sumaform and susemanager-ci') {
```
srbarrios, line 8: It seems it's not just cloning sumaform through terracumber, but also doing extra steps related to terraform files.
It might be good to clarify it on the title or even split the pure cloning of repos stage and other steps in different stages.

jenkins_pipelines/environments/common/pipeline-reference.groovy
```diff
@@ -0,0 +1,44 @@
...
+                // Create a directory for all the outputs belonging to this build number
+                sh "mkdir -p ${resultdir}"
+                git url: params.terracumber_gitrepo, branch: params.terracumber_ref
+                dir("susemanager-ci") {
+                    checkout scm
+                }
+                // Create a directory for  to place the directory with the build results (if it does not exist)
+                sh "mkdir -p ${resultdir}"
```
srbarrios, line 16: Why not use `dir` method as you did above? It will create the directory for you if it doesn't exist.

jenkins_pipelines/environments/common/pipeline-reference.groovy
```diff
@@ -0,0 +1,44 @@
...
+                sh "set +x; source /home/jenkins/.credentials set -x; ./terracumber-cli ${common_params} --gitrepo ${params.sumaform_gitrepo} --gitref ${params.sumaform_ref} --runstep gitsync"
+            }
+            stage('Deploy') {
+                // Provision the environment
+                if (params.terraform_bin_init) {
+                    env.TERRAFORM_INIT = '--init'
+                } else {
+                    env.TERRAFORM_INIT = ''
+                }
+                sh "set +x; source /home/jenkins/.credentials set -x; TERRAFORM=${params.terraform_bin} TERRAFORM_PLUGINS=${params.terraform_bin_plugins} ./terracumber-cli ${common_params} --logfile ${resultdirbuild}/sumaform.log ${env.TERRAFORM_INIT} --taint '.*(domain|main_disk).*' --runstep provision"
```
srbarrios, line 27: It might be good to validate that those Jenkins parameters exist. I guess we expect them in all our Jenkins job, but maybe checking the existence at the beginning can avoid some issues. Fine if we don't do anyway :)

jenkins_pipelines/environments/common/pipeline-reference.groovy
```diff
@@ -0,0 +1,44 @@
...
+                deployed = true
+            }
+        }
+        finally {
+            stage('Get results') {
+                if (!deployed) {
```
srbarrios, line 33: Going to the details, not asking to do, we will not support something that Jenkins uses to do, which is sent an email when the job came back to green, after a failed job.

jenkins_pipelines/environments/common/pipeline-reference.groovy
```diff
@@ -0,0 +1,44 @@
...
+                  // Send email
+                  sh "./terracumber-cli ${common_params} --logfile ${resultdirbuild}/mail.log --runstep mail"
+                }
+                // Clean up old results
+                sh "./clean-results -r ${resultdir}"
```
srbarrios, line 38: I'm not sure how this script behaves, it will do the clean up of the workspace?
We only want a clean up of the workspace if the pipeline had a fully success, as otherwise we might need some details in the workspace.
As we do here https://github.com/SUSE/susemanager-ci/blob/master/jenkins_pipelines/manager_testsuite/Jenkinsfile_qa_testsuite#L110

jenkins_pipelines/environments/common/pipeline.groovy
```diff
@@ -0,0 +1,86 @@
+def run(params) {
```
srbarrios, line 1: As we moved to those groovy scripts, it might be interesting to reuse some pieces of code like methods.
For instance here, I see that the first two stages are identical.

jenkins_pipelines/environments/common/pipeline.groovy
```diff
@@ -0,0 +1,86 @@
...
+                } else {
+                    env.TERRAFORM_INIT = ''
+                }
+                sh "set +x; source /home/jenkins/.credentials set -x; TERRAFORM=${params.terraform_bin} TERRAFORM_PLUGINS=${params.terraform_bin_plugins} ./terracumber-cli ${common_params} --logfile ${resultdirbuild}/sumaform.log ${env.TERRAFORM_INIT} --taint '.*(domain|main_disk).*' --runstep provision"
+                deployed = true
+            }
+            stage('Sanity Check') {
+                sh "./terracumber-cli ${common_params} --logfile ${resultdirbuild}/testsuite.log --runstep cucumber --cucumber-cmd 'cd /root/spacewalk/testsuite; rake cucumber:sanity_check'"
+            }
+
```
srbarrios, line 35: extra new line

jenkins_pipelines/environments/common/pipeline.groovy
```diff
@@ -0,0 +1,86 @@
...
+                                reportDir: "${resultdirbuild}/cucumber_report/",
+                                reportFiles: 'cucumber_report.html',
+                                reportName: "TestSuite Report"]
+                    )
+                    junit allowEmptyResults: true, testResults: "${junit_resultdir}/*.xml"
+                }
+                // Send email
+                sh "./terracumber-cli ${common_params} --logfile ${resultdirbuild}/mail.log --runstep mail"
+                // Clean up old results
+                sh "./clean-results -r ${resultdir}"
```
srbarrios, line 79: Same remark than before

jenkins_pipelines/environments/manager-3.2-reference-PRV
```diff
@@ -0,0 +1,26 @@
...
+        choice(name: 'terraform_bin', choices: ['/usr/bin/terraform_bin'], description: 'Terraform binary path'),
+        choice(name: 'terraform_bin_plugins', choices: ['/usr/bin'], description: 'Terraform plugins path'),
+        string(name: 'terracumber_gitrepo', defaultValue: 'https://gitlab.suse.de/juliogonzalezgil/terracumber.git', description: 'Terracumber Git Repository'),
+        string(name: 'terracumber_ref', defaultValue: 'master', description: 'Terracumber Git ref (branch, tag...)'),
+        booleanParam(name: 'terraform_bin_init', defaultValue: false, description: 'Call terraform_bin init (needed if modules are added or changes)')
+    ])
+])
+
+node('sumaform-cucumber') {
+    cron('H(0-30) 0-23/4 * * *')
```
srbarrios, line 20: Ref hosts might be better to be scheduled only nightly

jenkins_pipelines/environments/common/pipeline-reference.groovy
juliogonzalez: In fact I think the title is a typo. It should say "Clone terracumber" ;-)

Nice catch.

About the split: I thought about it. Problem is that I am not sure it adds value, and creates an extra step and makes the pipeline "bigger".

jenkins_pipelines/environments/common/pipeline-reference.groovy
juliogonzalez: Does it work even if I don't do anything else inside, such as `checkout scm`?

If so, I agree, I will use it.

jenkins_pipelines/environments/common/pipeline-reference.groovy
srbarrios: OK, fine with the renaming

jenkins_pipelines/environments/common/pipeline-reference.groovy
juliogonzalez: I think we better have default values. What do you think?

jenkins_pipelines/environments/common/pipeline-reference.groovy
juliogonzalez: That's right. We never supported it. For that we should use the Jenkins module. Still we should need this step in case we want to run from m226 if Jenkins is dead (as it happened in the past).

jenkins_pipelines/environments/common/pipeline-reference.groovy
juliogonzalez: No, this cleans the old results (so we only have the last 20 results). I think I am going to change the script name.

As far as I could see, no Jenkins job had that parameter `cleanWorkspace`, so that part never run.

But even so, that's dangerous. The tfstate is on the workspace, we should not clean it.

jenkins_pipelines/environments/common/pipeline.groovy
juliogonzalez: Absolutely right. I propose we evaluate it for a future PR (for now I wanted to avoid complicating the stuff for current pipeline developers, so we could stay as closer to the original as possible).

jenkins_pipelines/environments/common/pipeline.groovy
juliogonzalez: Will remove.

jenkins_pipelines/environments/common/pipeline.groovy
juliogonzalez: Same comment.

jenkins_pipelines/environments/common/pipeline-reference.groovy
srbarrios: Yes, it does, but why not add your line bellow inside that scope?
It makes total sense, as the command needs that directory, isn't it?

jenkins_pipelines/environments/common/pipeline-reference.groovy
srbarrios: Yep, but in that case, you don't have control of default values, as they are part of a different file, that's what I thought could be good to verify, but let's close the discussion, let's assume we don't fail :D

jenkins_pipelines/environments/common/pipeline-reference.groovy
juliogonzalez: In fact I was half wrong.

The correct title for this is `Clone terracumber, susemanager-ci and sumaform` as it's exactly what it does. First terracumber, then susemanager-ci (for the tf files) and then with both it clones/updates sumaform.

jenkins_pipelines/environments/common/pipeline-reference.groovy
srbarrios: Yep, you might be right, maybe we are doing wrong currently.

jenkins_pipelines/environments/common/pipeline-reference.groovy
juliogonzalez: Fixed. About using separate steps, to be clarified and if agreed to be done on a separate PR.

jenkins_pipelines/environments/common/pipeline-reference.groovy
juliogonzalez: Done, an inf fact this was duplicated (done also at line 10).

jenkins_pipelines/environments/common/pipeline-reference.groovy
juliogonzalez: Script renamed. Done.

jenkins_pipelines/environments/common/pipeline.groovy
juliogonzalez: Done

jenkins_pipelines/environments/common/pipeline-reference.groovy
juliogonzalez: Well, let's put it this way: the other file defines default values in case you want to have such parameters, for the builds.

But it could also be that most jobs would want to avoid having TERRAFORM and TERRAFORM_PLUGINS as well because they never ever change them.

That's why I think that in that case, we should define default values here.

jenkins_pipelines/environments/manager-3.2-reference-PRV
juliogonzalez: Ah, crap. Nice catch. Not "might be better", but "must be". This is a typo.

juliogonzalez: Fixed.

juliogonzalez:
Merging.

PR #76, opened 2020-05-08
Configure manager-Head-cucumber-NUE to use master again by default

juliogonzalez:


Bischoff:
Thanks, ping @rjmateus .

juliogonzalez:
@admd is aware, so merging.

PR #79, opened 2020-05-18
Renaming modules matching images used

srbarrios:
This PR fix the name of the module matching the image used, it's harmless, as in fact we are not using it anywhere.

Still, the attribute "name" must be updated, from sles15 to sles15sp1, but that will imply changes in the testsuite code too.


srbarrios:
Closing. It will not work. We need a refactor in sumaform.@rjmateus 
IMHO, we should change the required variables cli-sles12sp4 min-sles12sp4 minssh-sles12sp4 , by three variables (of type list, instead of a unique instance) of "client-list", "minion-list", "minionssh-list" or something similar..

rjmateus:
@srbarrios we have this issue opened https://github.com/SUSE/spacewalk/issues/10730

PR #81, opened 2020-05-22
Setup min-build VM

calancha:
Update terraform config. files to use the new min-build VM.
(This PR complements https://github.com/uyuni-project/uyuni/pull/2198)

### DIFF
--- a/terracumber_config/tf_files/SUSEManager-3.2-NUE.tf
+++ b/terracumber_config/tf_files/SUSEManager-3.2-NUE.tf
@@ -134,6 +134,11 @@ module "cucumber_testsuite" {
         mac = "AA:B2:93:00:21:68"
       }
     }
+    min-build = {
+      provider_settings = {
+        mac = "AA:B2:93:00:00:7B"
+      }
+    }
     minssh-sles12sp4 = {
       name = "minssh-sles12"
       provider_settings = {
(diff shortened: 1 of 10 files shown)


srbarrios:
Guessing you added the MAC address to the DHCP, and they are not colliding with others.
LGTM

srbarrios:
> > Guessing you added the MAC address to the DHCP, and they are not colliding with others.
> LGTM
> 
> And the DNS records as well.
> 
> Both are needed at gitlab's infrastructure repository.
> 
> If it's done, then PR LGTM.

As far as I remember, the MAC address are used to stick a MAC with a hostname, at DHCP level
But you don't set up MAC addresses in the DNS 🤔

juliogonzalez:
> As far as I remember, the MAC address are used to stick a MAC with a hostname, at DHCP level
> But you don't set up MAC addresses in the DNS thinking

You don't add MAC addresses to the DNS server. That's right.

But if you add a MAC address to DHCP, you will have a static IP, and if you have a static IP, probably you want a new DNS record for the instance on the environment, right?

So in general, if you add new instances, you should consider both DHCP and DNS.

juliogonzalez:
> As far as I remember, the MAC address are used to stick a MAC with a hostname

An IP, not a hostname.

calancha:
Pushing a commi to make the MAC's uniq.

calancha:
Squashing the 3 commits into 1.

Bischoff:
Changes at Provo dropped, let's start at NUE first.

PR #84, opened 2020-06-02
Add pipelines and tf files for SUSE Manager 4.1

juliogonzalez:


terracumber_config/tf_files/SUSEManager-4.1-NUE.tf
```diff
@@ -0,0 +1,183 @@
...
+      provider_settings = {
+        mac = "AA:B2:93:00:00:75"
+      }
+    }
+    pxy = {
+      provider_settings = {
+        mac = "AA:B2:93:00:00:81"
+      }
+    }
+    cli-sles12sp4 = {
```
juliogonzalez, line 125: We have sles12sp4 at Head and 4.0 as well, so I just did the same.

But why do we have it this way? Is it a sumaform limitation?

Bischoff: It's not sles12sp4 at all. It's a badly chosen name..

@rjmateus could we have this fixed once for good, and rename it e.g. `client` ?

terracumber_config/tf_files/SUSEManager-4.1-PRV.tf
```diff
@@ -0,0 +1,185 @@
...
+      provider_settings = {
+        mac = "52:54:00:00:00:29"
+      }
+    }
+  }
+  provider_settings = {
+    pool = "ssd"
+    network_name = null
+    bridge = "br0"
+    additional_network = "192.168.43.0/24"
```
Bischoff, line 179: The idea was to have private network IP ranges identical to the version. Was `192.168.41.0/24` impossible here?

terracumber_config/tf_files/SUSEManager-Head-NUE.tf
```diff
@@ -179,7 +179,7 @@ module "cucumber_testsuite" {
     pool = "ssd"
     network_name = null
     bridge = "br0"
-    additional_network = "192.168.41.0/24"
+    additional_network = "192.168.99.0/24"
```
juliogonzalez, line 182: As agreed with Eric, Uyuni uses `192.168.100.0/24`, so now head will use `192.168.99.0/24` so we don't need to change Head again.

After this is deployed, we'll need to taint head's extra network, or run a destroy.

terracumber_config/tf_files/SUSEManager-4.1-PRV.tf
juliogonzalez: Done.

terracumber_config/tf_files/SUSEManager-Head-NUE.tf
Bischoff: Thanks!

terracumber_config/tf_files/Uyuni-Master-NUE.tf
```diff
@@ -177,7 +177,7 @@ module "cucumber_testsuite" {
     pool               = "ssd"
     network_name       = null
     bridge             = "br0"
-    additional_network = "192.168.141.0/24"
+    additional_network = "192.168.100.0/24"
```
juliogonzalez, line 180: Adjusted Uyuni as well, as requested.

Bischoff: Thanks

terracumber_config/tf_files/SUSEManager-4.1-NUE.tf
rjmateus: @Bischoff we have a card to refactor it: https://github.com/SUSE/spacewalk/issues/10730
I can change just the names and them refactor the cucumber testsuite module. But now is not the time to do breaking changes, since we are in process of GA.

PR #85, opened 2020-06-05
Use sle15sp2 as a build host a retail terminal initial VM

lkotek:
We recently switched to SLES15SP2 as a retail terminal. This PR aligns OS used for the build host and initial OS for pxeboot VM.

### DIFF
--- a/terracumber_config/tf_files/SUSEManager-4.1-NUE.tf
+++ b/terracumber_config/tf_files/SUSEManager-4.1-NUE.tf
@@ -137,7 +137,7 @@ module "cucumber_testsuite" {
       }
     }
     min-build = {
-      image = "sles15sp1"
+      image = "sles15sp2"
       provider_settings = {
         mac = "AA:B2:93:00:00:84"
       }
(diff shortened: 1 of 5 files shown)


srbarrios:
Could you please try it from the branch before merge it? You can parametrize that in the jenkins  job.

lkotek:
> Could you please try it from the branch before merge it? You can parametrize that in the jenkins job.

Yes, I will definitely do so before merge.

lkotek:
> OK. There's still a bit of uncertainty whether this applies to 4.0 though. To be clarified first.
> 
> (if it doesn't, we'll have to split Kiwi profiles :-( )

This should be OK. I checked with the retail developers and code for this is the same in 4.0 and 4.1. So there shouldn't be any problem with building SLE15SP2 kiwi images on SLE15SP2 build host at SUMA 4.0 environment. Actually I tested exactly this on roc228 machine and it just worked :-) So we "only" needs confirmation there is no problem in case of terracumber (I hope not).

lkotek:
Jenkins CI were failing on unrelated error during night :-/ https://ci.suse.de/view/Manager/view/Manager-Head/job/manager-Head-cucumber-NUE/ So it wasn't possible to test my changes.

calancha:
This PR ssems necessary to fix recent failures at scenario `Check the OS image built as Kiwi image administrator`

PR #87, opened 2020-06-09
Use image sles15sp2 for min-build and proxy

calancha:
Use image sles15sp2 for both, min-build and pxeboot VM's at the 4.0 branch.

This completes 'use sle15sp2 as a build host a retail terminal initial vm'
(5640b5cc7d0150f96f5e383a2ae346fd8a2b9469)


calancha:
> Is it changed for NUE as well?
 `SUSEManager-4.0-NUE.tf` was covered by Lukas.
> What about ref40?
Neither `SUSEManager-4.0-refenv-NUE.tf`nor `SUSEManager-4.0-refenv-PRV.tf` were touched.
I am pushing a new commit to address these two files.

PR #88, opened 2020-06-30
Add openSUSE Leap 15.2 to Uyuni Master

juliogonzalez:
So we can move Server and Proxy to use Leap 15.2

That part is done at https://github.com/uyuni-project/sumaform/pull/723


Bischoff:
```
"opensuse150", "opensuse151", "opensuse152"
```

-> do we need to download that many images?

juliogonzalez:
Did we already migrate all the stuff to at least 15.1?

If so, whoever did it needs to take care of the cleanup, because we have `opensuse150` at all TF files :-)

But that would be a separate PR.

Bischoff:
15.1 is still needed for KVM build host, and 15.0 for controller (I know, not supported anymore...)

ok to leave as is then.

PR #90, opened 2020-07-09
Change 4.1 and Head to use SLES15SP2 for the KVM minion

juliogonzalez:
Reason: because I can't test Leap 15.2 and KVM minion on Uyuni (we don't inject salt packages there, and Leap 15.2 doesn't have the correct salt 3000 yet, that wil come after 4.0.7 release).

So at least this way we can validate KVM on SLES15SP2.


juliogonzalez:
Ping @cbosdo for review (can't add you for some reason)

PR #96, opened 2020-07-14
Adjust mac addresses to DHCP/DNS changes requried to add suma-41-srv

juliogonzalez:
It requires the PRs for DNS/DHCP merged in advance.

In the end the srv mac address for 4.1-PRV was right, but wasn't added to DNS/DNCP (the PRs do that). And to have everything right, the IPs and MACs for ref41 where bumped by adding `1`.

### DIFF
--- a/terracumber_config/tf_files/SUSEManager-4.1-PRV.tf
+++ b/terracumber_config/tf_files/SUSEManager-4.1-PRV.tf
@@ -135,7 +135,7 @@ module "cucumber_testsuite" {
       image = "sles15sp1"
       name = "min-sles15"
       provider_settings = {
-        mac = "52:54:00:00:00:33"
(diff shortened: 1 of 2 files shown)


terracumber_config/tf_files/SUSEManager-4.1-PRV.tf
```diff
@@ -135,7 +135,7 @@ module "cucumber_testsuite" {
       image = "sles15sp1"
       name = "min-sles15"
       provider_settings = {
-        mac = "52:54:00:00:00:33"
```
juliogonzalez, line 138: This mac address was wrong, as `52:54:00:00:00:33` was for `suma-ref41-min-sles15` (and now for `suma-ref41-cli-sles15`)

PR #98, opened 2020-07-22
Add a 60 minutes timeout for the testsuites

juliogonzalez:
The `activity` parameter is used so the timeout applies to "time without activity on the logs":

https://www.jenkins.io/doc/pipeline/steps/workflow-basic-steps/#timeout-enforce-time-limit

We could adjust it individually for each job if needed.

If you think more than 60 minutes are needed (or less), please do let me know.

I tested https://github.com/juliogonzalez/jenkins-pipeline-parameters (changing the value for `time` at `product3.2`) at my private Jenkins, and worked fine.

### DIFF
--- a/jenkins_pipelines/environments/manager-3.2-cucumber-NUE
+++ b/jenkins_pipelines/environments/manager-3.2-cucumber-NUE
@@ -23,6 +23,8 @@ node('sumaform-cucumber') {
     stage('Checkout pipeline') {
         checkout scm
     }
-    def pipeline = load "jenkins_pipelines/environments/common/pipeline.groovy"
-    pipeline.run(params)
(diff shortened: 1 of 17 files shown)


juliogonzalez:
> About the value you chose (60 minutes) : I think we had a bit more (90 minutes) in the past, but I might be remembering wrong.

I think you are right, because of some docker builds, maybe?

I'd say: let's try with 60 minutes, maybe things are better now, and if it still fails, it's easy to change.

juliogonzalez:
For some reason this is not working exactly as it should (sometimes the builds get terminated even with activity in the last 60 minutes), and no time to debug it now before going on holidays. Since I need results ASAP, I am going to switch it to a 10 hours absolute timeout with a new PR.

PR #101, opened 2020-07-24
Add the SLE15SP2 development tools for the 4.1 refserver until 4.1.1 is out

juliogonzalez:


terracumber_config/tf_files/SUSEManager-4.1-refenv-PRV.tf
```diff
@@ -106,6 +106,10 @@ module "srv" {
   provider_settings = {
     mac    = "52:54:00:00:00:32"
     memory = 8192
+    additional_repos = {
```
rjmateus, line 109: this should be placed outside the provider settings

juliogonzalez: You right. Wrong copy paste.

PR #102, opened 2020-07-29
Use better name for hosts in testsuite module

Bischoff:
sumaform: uyuni-project/sumaform#734
terracumber: https://gitlab.suse.de/juliogonzalezgil/terracumber/-/merge_requests/7

### DIFF
--- a/jenkins_pipelines/environments/manager-3.2-cucumber-NUE
+++ b/jenkins_pipelines/environments/manager-3.2-cucumber-NUE
@@ -14,8 +14,9 @@ node('sumaform-cucumber') {
             choice(name: 'sumaform_backend', choices: ['libvirt', 'aws'], description: 'Sumaform backend to be used (see https://github.com/uyuni-project/sumaform#backend-choice)'),
             choice(name: 'terraform_bin', choices: ['/usr/bin/terraform_bin'], description: 'Terraform binary path'),
             choice(name: 'terraform_bin_plugins', choices: ['/usr/bin'], description: 'Terraform plugins path'),
-            string(name: 'terracumber_gitrepo', defaultValue: 'https://gitlab.suse.de/juliogonzalezgil/terracumber.git', description: 'Terracumber Git Repository'),
-            string(name: 'terracumber_ref', defaultValue: 'master', description: 'Terracumber Git ref (branch, tag...)'),
+            // Temporary: should move to uyuni-project
+            string(name: 'terracumber_gitrepo', defaultValue: 'https://gitlab.suse.de/ebischoff/terracumber.git', description: 'Terracumber Git Repository'),
+            string(name: 'terracumber_ref', defaultValue: 'multiple-hosts-testsuite', description: 'Terracumber Git ref (branch, tag...)'),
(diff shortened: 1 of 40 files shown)


Bischoff:
> Generally - I like the changes :-) I am only thinking about using "redhat" as a part of the name. What about using something like "rhel-family" instead of it? I understand it is longer, but it also covers cases when we use Centos etc.

"RHEL" stands for "RedHat Enterprise Linux" and it's a specific product from RedHat.

Yes, what is implied there is a "family" of products. For example "suse" family includes "openSUSE" product even though it's not controlled by SUSE as a company. If any, it would then be "redhat-family" and would include CentOS, Fedora, SLES ES, etc. I was tempted to add "family" to make it explicit but I was afraid it was becoming a bit too long.

lkotek:
> > Generally - I like the changes :-) I am only thinking about using "redhat" as a part of the name. What about using something like "rhel-family" instead of it? I understand it is longer, but it also covers cases when we use Centos etc.
> 
> "RHEL" stands for "RedHat Enterprise Linux" and it's a specific product from RedHat.
> 
> Yes, what is implied there is a "family" of products. For example "suse" family includes "openSUSE" product even though it's not controlled by SUSE as a company. If any, it would then be "redhat-family" and would include CentOS, Fedora, SLES ES, etc. I was tempted to add "family" to make it explicit but I was afraid it was becoming a bit too long.

OK, I understand your point. Let's keep it as it is then :)

Bischoff:
> OK, I understand your point. Let's keep it as it is then :)

Oscar seems to share your view, based on `ID_LIKE` field of https://www.freedesktop.org/software/systemd/man/os-release.html .

I can follow the majority and rename `redhat`to `rhel` 😸  .

Bischoff:
Changed my mind:

Salt uses "Debian" and "RedHat" as families 🙂
```
{% elif grains['os_family'] == 'RedHat' %}
```

srbarrios:
> Changed my mind:
> 
> Salt uses "Debian" and "RedHat" as families 🙂
> 
> ```
> {% elif grains['os_family'] == 'RedHat' %}
> ```

To be concrete, the salt states on sunaform use this text as family ;)

Bischoff:
> > Changed my mind:
> > ```
> > {% elif grains['os_family'] == 'RedHat' %}
> > ```
> 
> To be concrete, the salt states on sunaform use this text as family ;)

OK. But at least we're compatible with existing sumaform...

juliogonzalez:
https://gitlab.suse.de/juliogonzalezgil/terracumber/-/merge_requests/7 is merged, so the original repo can be used again (moving to Uyuni will take some time, there are some final changes I want to do first)

Bischoff:
> https://gitlab.suse.de/juliogonzalezgil/terracumber/-/merge_requests/7 is merged, so the original repo can be used again

Cool @juliogonzalez . I'm on vacation but I will create the PR now, it's the matter of 10 minutes.

> (moving to Uyuni will take some time, there are some final changes I want to do first)

No worries.

Bischoff:
Back to your repository in #102.

PR #104, opened 2020-08-10
Qam new images

ktsamis:
New images centos 6, centos 7, centos 8, ubuntu 16, ubuntu 18, ubuntu 20. Some of them are new products, others needed fixing because they were commented out and obsolete settings. My recommendation is to just try to deploy this from my fork and see if it works. I haven't tested it because I don't have the computing power.

Fixes https://github.com/SUSE/spacewalk/issues/9360

PR #105, opened 2020-08-11
Switch to 4.1-released the testsuite

ktsamis:
https://github.com/SUSE/susemanager-ci/pull/104#pullrequestreview-464866181
Time to switch I think to released. Obviously not tested this, we have to see the next results

### DIFF
--- a/terracumber_config/tf_files/SUSEManager-4.1-NUE.tf
+++ b/terracumber_config/tf_files/SUSEManager-4.1-NUE.tf
@@ -81,8 +81,8 @@ provider "libvirt" {
 module "cucumber_testsuite" {
   source = "./modules/cucumber_testsuite"
 
-  product_version = "4.1-nightly"
-  
+  product_version = "4.1-released"
+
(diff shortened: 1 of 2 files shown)


terracumber_config/tf_files/SUSEManager-4.1-NUE.tf
```diff
@@ -81,8 +81,8 @@ provider "libvirt" {
 module "cucumber_testsuite" {
   source = "./modules/cucumber_testsuite"
 
-  product_version = "4.1-nightly"
-  
+  product_version = "4.1-released"
```
juliogonzalez, line 84: I don't really understand why this PR is changing this. The two TFs are used to test Devel:Galaxy:Manager:4.1*

If you change it to released, you will test the released code.

ktsamis:
We don't want to have released for dev code, we want nightly, wrong PR

Bischoff:
Yes, because we want to test upcoming MUs.

PR #106, opened 2020-08-14
Use official images in all CI

calancha:
Update all main.tf to download and use official images by default.

We have merged https://github.com/uyuni-project/sumaform/pull/741 to
use official images by default in sumaform.   Now we need to update the main.tf
so that the default images are available.

- SUSEManager-4.1-NUE.tf:
Add the additional repos in the server

### DIFF
--- a/terracumber_config/tf_files/SUSEManager-3.2-refenv-PRV.tf
+++ b/terracumber_config/tf_files/SUSEManager-3.2-refenv-PRV.tf
@@ -81,7 +81,7 @@ module "base" {
   name_prefix       = "suma-ref32-"
   use_avahi         = false
   domain            = "prv.suse.net"
-  images            = ["centos7", "sles12sp4", "ubuntu1804"]
+  images            = ["centos7o", "sles12sp4o", "ubuntu1804o"]
   mirror            = "minima-mirror.prv.suse.net"
   use_mirror_images = true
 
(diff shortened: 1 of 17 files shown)


terracumber_config/tf_files/SUSEManager-3.2-refenv-PRV.tf
```diff
@@ -118,7 +118,7 @@ module "suse-client" {
   base_configuration = module.base.configuration
   product_version    = "3.2-nightly"
   name               = "cli-sles12"
-  image              = "sles12sp4"
+  image              = "sles12sp4o"
```
juliogonzalez, line 121: If you say we changed the default images at https://github.com/uyuni-project/sumaform/pull/741, why do we need to specify the images here? (same for all other places)

calancha: You are right.  We don't need tuch an eager commit; we can keep the `image` variables in those files.
Then, the only we need to do here is to update the variable `images`.
- it must contain defaults as centos7o or sles15sp1o in those main.tf where the default is implied
(e.g.  `SUSEManager-4.1-PRV.tf`).

calancha: Now, `image` is only specified when it sets a value other than the default.

calancha:
> There's something that maybe is right, but I don't understand.

What is it?

calancha:
I am going to squash the commits and then merge: the test suites are failing because they need these changes.

PR #107, opened 2020-08-14
Do not use sles15sp1o

calancha:
## What this PR change

Do not use sles15sp1o image.
This image is causing problems: we are not able to
deploy when this image is used.


calancha:
We have solved all issues with this image; we are using it now.

PR #114, opened 2020-09-03
Official images again in 4.0 NUE

Bischoff:
Use -o images again in 4.0 NUE

Note: if we prefer, we can close this, and do a full revert of the revert of the official images (i.e. all branches at once)


Bischoff:
Merging, otherwise my changes to master are not used

PR #115, opened 2020-09-04
Test salt authentication

admd:
Change config to run cucumbers on correct branches.

### DIFF
--- a/terracumber_config/tf_files/SUSEManager-Test-NUE.tf
+++ b/terracumber_config/tf_files/SUSEManager-Test-NUE.tf
@@ -12,12 +12,12 @@ variable "CUCUMBER_COMMAND" {
 
 variable "CUCUMBER_GITREPO" {
   type = "string"
-  default = "https://github.com/uyuni-project/uyuni.git"
+  default = "https://github.com/SUSE/spacewalk.git"
 }
 
 variable "CUCUMBER_BRANCH" {
   type = "string"
-  default = "master"
+  default = "Manager-4.0"
 }
 
 variable "CUCUMBER_RESULTS" {
@@ -81,8 +81,8 @@ provider "libvirt" {
 module "cucumber_testsuite" {
   source = "./modules/cucumber_testsuite"
 
-  product_version = "head"
-  
+  product_version = "4.0-released"
+
   // Cucumber repository configuration for the controller
   git_username = var.GIT_USER
   git_password = var.GIT_PASSWORD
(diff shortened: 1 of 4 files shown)


terracumber_config/tf_files/SUSEManager-Test-NUE.tf
```diff
@@ -81,8 +81,8 @@ provider "libvirt" {
 module "cucumber_testsuite" {
   source = "./modules/cucumber_testsuite"
 
-  product_version = "head"
-  
+  product_version = "4.0-released"
```
juliogonzalez, line 84: You are mixing: `Manager-4.1` branch for cucumber, and `4.0-released` for the product version.

But besides that: are you sure 1000% sure this should be `released`?

terracumber_config/tf_files/SUSEManager-Test-Naica-NUE.tf
```diff
@@ -12,12 +12,12 @@ variable "CUCUMBER_COMMAND" {
...
 variable "CUCUMBER_GITREPO" {
   type = "string"
-  default = "https://github.com/uyuni-project/uyuni.git"
+  default = "https://github.com/SUSE/spacewalk.git"
 }
 
 variable "CUCUMBER_BRANCH" {
   type = "string"
-  default = "master"
+  default = "Manager-3.2"
```
juliogonzalez, line 20: Same

terracumber_config/tf_files/SUSEManager-Test-Naica-NUE.tf
```diff
@@ -81,8 +81,8 @@ provider "libvirt" {
 module "cucumber_testsuite" {
   source = "./modules/cucumber_testsuite"
 
-  product_version = "head"
-  
+  product_version = "3.2-released"
```
juliogonzalez, line 84: Same as before.

juliogonzalez (changes requested):
Some changes required.

And... aren't we missing a change to test 4.1?

terracumber_config/tf_files/SUSEManager-Test-NUE.tf
admd: ah right, 4.1 was a typo, thank you.

I don't see any problem with released as I am testing a branch based on the last released MU. Do you see any problem?

admd:
> And... aren't we missing a change to test 4.1?

For 4.1, we already have changes in place against Hexagon

terracumber_config/tf_files/SUSEManager-Test-NUE.tf
juliogonzalez: I am pretty sure that if you use it, you will only get `released` packages, without the packages you want to test. Check the sumaform doc.

juliogonzalez (approved):
Did not notice any error, so let's give it a try.

PR #117, opened 2020-09-07
Mount the extra mirrors from the server

Bischoff:
For QAM testsuite:

* use second mirror (we might revert it if we use the big mirror on ceph cluster offered by paul)
* fix private network


srbarrios:
These changes are supported by sumaform master branch?
If so, we can merge :)

Bischoff:
Yes, that's the point@obarrios: they don't rely on sumaform to work, just some `main.tf` magic.

PR #118, opened 2020-09-09
Use hexagon environment to test Uyuni 2020.07 fixes

juliogonzalez:


moio (changes requested):
I think this line needs changing too 

https://github.com/SUSE/susemanager-ci/pull/118/files#diff-3f5d9dc28a58069171660f28ab314e18R84

juliogonzalez:
True! Fixed now :-)

moio (commented):
Just noticed, the image list might also need tweaking, as opensuse152o is not there

juliogonzalez:
Added

moio (approved):

PR #119, opened 2020-09-10
Let openscap run with more RAM on Centos minion on 4.0 and 4.1

Bischoff:
With only one GB on CentOS for 4.0 and 4.1:
```
[root@suma-40-min-centos7 ~]# oscap xccdf eval --profile standard /usr/share/xml/scap/ssg/content/ssg-centos7-xccdf.xml >/dev/null
WARNING: This content points out to the remote resources. Use `--fetch-remote-resources' option to download them.
WARNING: Skipping https://www.redhat.com/security/data/oval/com.redhat.rhsa-RHEL7.xml file which is referenced from XCCDF content
W: oscap:         Can't connect: 12, Cannot allocate memory.
E: oscap:         Connect: retry limit (0) reached.
W: oscap:       Can't connect: 12, Cannot allocate memory.
E: oscap:       Connect: retry limit (0) reached.
W: oscap:       Can't connect: 12, Cannot allocate memory.
E: oscap:       Connect: retry limit (0) reached.
W: oscap:       Can't connect: 12, Cannot allocate memory.
E: oscap:       Connect: retry limit (0) reached.
```
etc

### DIFF
--- a/terracumber_config/tf_files/SUSEManager-4.0-NUE.tf
+++ b/terracumber_config/tf_files/SUSEManager-4.0-NUE.tf
@@ -152,6 +152,8 @@ module "cucumber_testsuite" {
     redhat-minion = {
       provider_settings = {
         mac = "AA:B2:93:00:00:44"
+        // Openscap cannot run with less than 1.25 GB of RAM
+        memory = 1280
       }
     }
     debian-minion = {
--- a/terracumber_config/tf_files/SUSEManager-4.0-PRV.tf
+++ b/terracumber_config/tf_files/SUSEManager-4.0-PRV.tf
@@ -154,6 +154,8 @@ module "cucumber_testsuite" {
     redhat-minion = {
       provider_settings = {
         mac = "52:54:00:00:00:05"
+        // Openscap cannot run with less than 1.25 GB of RAM
+        memory = 1280
       }
     }
     debian-minion = {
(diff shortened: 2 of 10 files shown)


juliogonzalez (commented):
What about 4.0, 3.2 and the test environments?

juliogonzalez (approved):
If they are already correct, then this LGTM

Bischoff:
> What about 4.0, 3.2 and the test environments?

To be very honest, I don't know. But I prefer not touching them if I have no proof that they are also affected.

According to the old comment, the reboot problem affected only head at that time.

I will retest the openscap problem right now on 3.2 and 4.0.

srbarrios (approved):

juliogonzalez:
> To be very honest, I don't know. But I prefer not touching them if I have no proof that they are also affected.

Makes sense :+1:

calancha (approved):

Bischoff:
> I will retest the openscap problem right now on 3.2 and 4.0.

Soooo it's a big mess. From what I can see:

* 3.2 - openscap OK  reboot OK
* 4.0 - openscap NOT OK   reboot OK
* 4.1 - openscap NOT OK   reboot OK

Is the reboot problem gone ? Or am I testing it wrong? I'll change the PR accordingly.

PR #130, opened 2020-09-30
Use new network for 4.1 PRV

Bischoff:
Use new DHCP and DNS for 4.1 Provo test suite.

### DIFF
--- a/terracumber_config/tf_files/SUSEManager-4.1-PRV.tf
+++ b/terracumber_config/tf_files/SUSEManager-4.1-PRV.tf
@@ -97,7 +97,7 @@ module "cucumber_testsuite" {
 
   use_avahi = false
   name_prefix = "suma-41-"
-  domain = "prv.suse.net"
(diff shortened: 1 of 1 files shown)


juliogonzalez (approved):

srbarrios:
What's the reason behind the need of change the bridge to br1?
Now we have two bridges? This is just temporary?

Bischoff:
> What's the reason behind the need of change the bridge to br1?
> Now we have two bridges? This is just temporary?
* br0 is connected to vlan 1075; on this vlan, the dhcp server is the one from SCC team and creates IP addresses from prv.suse.net
* br1 is connected to vlan 1356; on this vlan, the dhcp server is the one I prepared and creates IP addresses from mgr.prv.suse.net

by having two bridges, we can have a (hopefully) smooth transitions, with some machines still in prv.suse.net, while others are already in mgr.prv.suse.net

PR #139, opened 2020-10-13
Remove Oracle, as 3.2 is EoL

juliogonzalez:
After we merge, I suggest we create a tag 'oracle-removal'.

We can't still remove 3.2, so this PR is not about removing 3.2.


Bischoff (approved):
LGTM

mcalmer (approved):

PR #146, opened 2020-10-22
Avoid false duplicate MAC addresses

Bischoff:
This PR prevents MAC address duplicates with private network by using a different prefix (`aa:b2:92` instead of `52:54:00`).

It also fixes one copy/paste error in QAM test suite (`ubuntu2004-sshminion` using `52:54:00:2A:47:D8` instead of `52:54:00:0E:28:39`).

See also https://gitlab.suse.de/galaxy/infrastructure/-/merge_requests/203


srbarrios (commented):
LGTM

PR #152, opened 2020-11-02
Upgrade the opensuse image for the controller module

srbarrios:
This PR upgrades the image used on the controller module. It will do it for all our environments.
At same time, we will change sumaform's default values, to use the new version of Leap as image for the controller module.

WARNING: This PR must be merged together with uyuni-project/sumaform#790


calancha:
Shouldn't `SUSEManager-4.0-PRV.tf` use `opensuse152o` as well?

srbarrios:
> Shouldn't `SUSEManager-4.0-PRV.tf` use `opensuse152o` as well?

Good catch fixed.

calancha (approved):
LGTM.  Thank you.

PR #161, opened 2020-11-11
In the case of QAM we have more resources to taint

Bischoff:
Taint also:

* the data disk for /var/spacewalk
* the extra NFS mounts from mirror2


srbarrios (approved):

ktsamis (approved):
LGTM

PR #166, opened 2020-11-18
Move internal registry to registry.mgr.prv.suse.net

jordimassaguerpla:
partial fix for https://github.com/SUSE/spacewalk/issues/13060


Bischoff (changes requested):
I would try first on only one machine, for example 4.0 provo

We will need 4.1 provo soon so better not touch it until we are sure it works

PR #170, opened 2020-12-17
QA: Support run a testsuite filtering tests by scope

srbarrios:
Second part of https://github.com/SUSE/spacewalk/issues/12831

By using the environment variable PROFILE, we provide via Jenkins choice parameter, a Cucumber profile to our Rake task.
This will allow us to run a test suite running only a certain profile, which normally matches a certain group.

Note that this is the first approach. If people embrace it, we can re-visit and we might change it from a choice parameter to checkboxes, so we can select more than one profile. But that will also require a change in our product code.


Bischoff (approved):
LGTM

PR #172, opened 2021-01-07
QAM: Adding VMs necessary for retail tests

lkotek:
We need 3 buildhosts for building all types of retail kiwi images and the same number of retail terminals. We also need to assure private network is located at the same hypervisor as proxy (branch server) and all the terminals for pxe booting.

This PR must be merged after: https://gitlab.suse.de/galaxy/infrastructure/-/merge_requests/261

### DIFF
--- a/terracumber_config/tf_files/SUSEManager-4.1-qam.tf
+++ b/terracumber_config/tf_files/SUSEManager-4.1-qam.tf
@@ -132,7 +132,6 @@ module "base_old_sle" {
   provider_settings = {
     pool        = "default"
     bridge      = "br1"
-    additional_network = "192.168.41.0/24"
   }
 }
 
@@ -183,7 +182,6 @@ module "base_newsle_ubuntu" {
   provider_settings = {
     pool        = "default"
     bridge      = "br1"
-    additional_network = "192.168.41.0/24"
   }
 }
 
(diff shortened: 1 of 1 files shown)


srbarrios (approved):
LGTM.
Just please check if the sum of all the vCPUs and RAM in the different hypervisors still not overcome the maximum value of the harware. Otherwise, we might have some raise conditions overloading the hypervisor's limits.

terracumber_config/tf_files/SUSEManager-4.1-qam.tf
```diff
@@ -866,6 +864,135 @@ module "ubuntu1604-sshminion" {
...
+  providers = {
+    libvirt = libvirt.giediprime
+  }
+  source             = "./modules/minion"
+  base_configuration = module.base_old_sle.configuration
+  product_version    = "4.1-released"
+  name               = "buildhost-sles11sp4"
+  image              = "sles11sp4"
+  provider_settings = {
+    mac                = "aa:b2:92:6a:52:83"
```
calancha, line 877: Two of the MACs are already assigned:
```
suma-qam-41-min-sles11sp4: aa:b2:92:6a:52:83
suma-qam-41-min-sles12sp4: aa:b2:92:9a:94:c9
```

Shouldn't we add new uniq MACs for these added VMs?

lkotek: Good catch, I already prepared new mac addresses to be used in the second PR for infrastructure, but I forgot to paste them here. Thanks a lot, I will fix it asap :-)

lkotek:
> LGTM.
> Just please check if the sum of all the vCPUs and RAM in the different hypervisors still not overcome the maximum value of the harware. Otherwise, we might have some raise conditions overloading the hypervisor's limits.

Unfortunately we do. We already use 14 vCPUs, but there are only 12 of them according to `cpuinfo` at `arrakis`. And VMs in my PR need 4 more vCPUs (also these VMs must be located at same hypervisor where proxy is located).

Can we afford this?

Only good thing here I see is that these VMs will be used for relatively short amount of time when terminals are deployed and booted.

srbarrios:
We can try, but I think that terraform will fail of we try to allocate more vCPUs than the hypervisor has.

Ideally, we should move some VMs to a fourth hypervisor.

PR #174, opened 2021-01-12
Enable SP migration at reposync and init_clients stages

calancha:
In order to run the SP migration by default, we must
inject the environment variable controlling that into the pipe.

Add parameter service_pack_migration with default true.
Export SERVICE_PACK_MIGRATION in all the stages that need this variable.
- jenkins_pipelines/environments/common/pipeline.groovy
- jenkins_pipelines/environments/manager-4.1-cucumber-NUE
- jenkins_pipelines/environments/manager-4.1-cucumber-PRV
- jenkins_pipelines/environments/manager-Head-cucumber-NUE
- jenkins_pipelines/environments/manager-Head-reference-NUE
- jenkins_pipelines/environments/uyuni-master-cucumber-NUE


srbarrios (approved):

srbarrios:
Approved. We will refactor it in a future PR, so we handle this variable in the negative case, to handle the tag on scenario scope instead of feature scope.

Warning, if we unmark this new checkbox, the feature will fail, as no minion will be boostrapped.

PR #175, opened 2021-01-14
Enable 15sp2 in qam

ktsamis:
Just removing the comment lines


ktsamis:
seems at the same time we enabled. closing

PR #183, opened 2021-02-01
Add custom grain build_validation

ktsamis:
Add custom grain build_validation to be read by sumaform. Fixes https://github.com/SUSE/spacewalk/issues/13849

### DIFF
--- a/terracumber_config/tf_files/SUSEManager-4.1-qam.tf
+++ b/terracumber_config/tf_files/SUSEManager-4.1-qam.tf
@@ -1110,6 +1110,7 @@ module "controller" {
   git_repo     = var.CUCUMBER_GITREPO
   branch       = var.CUCUMBER_BRANCH
 
+  grains = { build_validation = "true" }
   server_configuration = module.server.configuration
   proxy_configuration  = module.proxy.configuration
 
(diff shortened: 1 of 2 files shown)


ktsamis:
@rjmateus Please check that the definition of the custom grain and the reading of it is fine.

terracumber_config/tf_files/SUSEManager-4.1-qam.tf
```diff
@@ -1110,6 +1110,7 @@ module "controller" {
   git_repo     = var.CUCUMBER_GITREPO
   branch       = var.CUCUMBER_BRANCH
 
+  grains = { build_validation = "true" }
```
srbarrios, line 1113: I think we are over-engineering here. I would really avoid any extra parameter in the main.tf or sumaform.
We can just inject this variable from the pipeline, using terracumber.

Bischoff: No, no, no :-) . Please.

I want to be able to deploy myself (no jenkins) => sumaform.

And even when deployed by Jenkins, when I test manually, I do not want to need to do export BUILD_VALIDATION=true by hand every time i test.

Bischoff (approved):
LGTM

ktsamis:
After discussions we will remove this completely. not needed anymore

PR #184, opened 2021-02-01
Permanent Build Validation variable

srbarrios:
Harder to explain that to write , so I wrote my approach to have permanently BUILD_VALIDATION variable in the controller.


srbarrios:
Closing.
What we agreed to do is to get rid of this BUILD_VALIDATION variable.
We don't need it anymore, it was a bad small step done, when designing build validation pipeline, where I though we don't use custom repositories. Now that we clarified this point, having a custom_repositories.json in QAM and Build Validaitons, we can have the same approach we had before, so using this json file to identify this use case.

PR #187, opened 2021-02-15
Fix 15sp3 name and image for sshminion for new hypervisors main.tf

ktsamis:
Basically a port for https://github.com/SUSE/susemanager-ci/pull/186 for the other main.tf


ktsamis:
merged because this is a port essentially

PR #190, opened 2021-02-19
Add debian minions and sshminions in advance

ktsamis:
Add debian minions and sshminions in advance, they will be commented out even when this is merged.

PR #191, opened 2021-02-22
Rename base_newsle_ubuntu to base_newsle_debian

ktsamis:
Rename base_newsle_ubuntu to base_newsle_debian from comment https://github.com/SUSE/susemanager-ci/pull/190#discussion_r580162120

### DIFF
--- a/terracumber_config/tf_files/SUSEManager-4.0-qam.tf
+++ b/terracumber_config/tf_files/SUSEManager-4.0-qam.tf
@@ -165,7 +165,7 @@ module "base_res" {
   }
 }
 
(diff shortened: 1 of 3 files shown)


srbarrios (approved):

Bischoff:
Yeah, it was a very sheldonish comment of mine.

But we are also moving to "debian" for general case ubuntu / debian in sumaform, so let's go that path.

Bischoff (approved):
Yeah, it was a very sheldonish comment of mine.

But we are also moving to "debian" for general case ubuntu / debian in sumaform, so let's go that path.

PR #192, opened 2021-02-22
BV: Use the new pipeline not only for 4.2 but also for 4.1.x

srbarrios:
- Remove old pipelines for QAM 4.0
- Replace old pipelines for QAM 4.1 for a new pipeline for BV 4.1


ktsamis (approved):
Let's get rid of the ones we don't use, looks great

PR #194, opened 2021-02-24
Sync SLE15SP1 channels for 4.0 reference environments

juliogonzalez:
We have SLE15SP1 client, so we need SLE15SP1 channels (same as we do for 4.1).

This starting failing since I removed SLE15SP0 from the mirror on December, and seems nobody complained, so looks to me that the ref40 environment is not that popular :-)


srbarrios (approved):
Just a doubt, we also need here the LTSS repos?

juliogonzalez:
> Just a doubt, we also need here the LTSS repos?

I'd say no, because we don't have any LTSS clients. What we could need (in the future) is SP2, because the buildhost is a SP2 image. That would be needed for 4.1 refenvs as well (and maybe Head?)

But for now this will at least fix the deployment, that otherwise fails.

PR #196, opened 2021-03-08
Split new SLE and Debian on 2 hypervisors

Bischoff:
This PR uses the new "mandalore" hypervisor for Ubuntu VMs.

It also fixes a few errors:
* 4.1: missing `server_mounted_mirror = "minima-mirror-qam.mgr.prv.suse.net"` declaration
* 4.2:  `base_core.configuration` used for proxy instead of `base_retail.configuration`
* 4.2: `//sle12sp4-minion_additional_repos` used for sle12sp5

Finally it converges 4.1 and 4.2 main.tf's.

### DIFF
--- a/terracumber_config/tf_files/SUSEManager-4.1-build-validation.tf
+++ b/terracumber_config/tf_files/SUSEManager-4.1-build-validation.tf
@@ -170,7 +170,7 @@ module "base_res" {
   }
 }
 
-module "base_newsle_debian" {
+module "base_new_sle" {
   providers = {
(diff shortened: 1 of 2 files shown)


ktsamis (approved):

Bischoff:
@lkotek @srbarrios you are not here today - I am fearing that some of my "error fixes" might be incorrect, but you are not here today to review.

So here is what we will do: I will merge, and if there is a problem with one or more of these "fixes", we will just revert them afterwards.

PR #197, opened 2021-03-08
Change description of cucumber_ref value on 4.2 BV and remove whitespace

ktsamis:
Change description of cucumber_ref value on 4.2 BV and remove whitespace


Bischoff (approved):
lgtm

PR #206, opened 2021-03-17
Enable ubuntu20 and debian9 and debian10 minions in BV

ktsamis:
Enable ubuntu20 and debian9 and debian10 minions in BV


ktsamis:
Made this WIP just to merge whenever we want to fix deployment and then use them.

Bischoff (approved):

PR #222, opened 2021-04-01
Add request review and author assign workflows

ktsamis:
Add request review and author assign workflows, I put some example files in reviewers.yml , we can add more or subtract. Also we should add each squad as reviewers for their own tf files to be updated every time they change. Or not. Let's discuss it.

### DIFF
--- a/.github/reviewers.yml
+++ b/.github/reviewers.yml
@@ -1,0 +1,26 @@
+reviewers:
+  # Reviewer groups each of which has a list of GitHub usernames
+  groups:
+    qa:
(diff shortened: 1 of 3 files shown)


srbarrios:
I agree on having it in spacewalk and uyuni, but here.. in our pipelines... maybe it's too much, I personally like some freedom on this repo. But ok, I will not oppose.

ktsamis:
So the auto assign author we removed already from uyuni. And for reviews for this repo maybe is too much. I think I will close this PR without merging

PR #247, opened 2021-04-22
Add new experimental java in all CI runs to test

ktsamis:
Fixes https://github.com/SUSE/spacewalk/issues/14634

It might cause problems in the build validation repos, we might have duplicate the additional_repos key. I will see in the next deployment and fix it. The fix would be simply to remove it from main.tf and add it in the default template in the json

### DIFF
--- a/terracumber_config/tf_files/Uyuni-Master-NUE.tf
+++ b/terracumber_config/tf_files/Uyuni-Master-NUE.tf
@@ -82,7 +82,7 @@ module "cucumber_testsuite" {
   source = "./modules/cucumber_testsuite"
 
   product_version = "uyuni-master"
-  
+
   // Cucumber repository configuration for the controller
   git_username = var.GIT_USER
   git_password = var.GIT_PASSWORD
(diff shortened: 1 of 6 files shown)


ktsamis:
@juliogonzalez Can you please check that the repo is the correct one? It does correspond to the OBS project but could it be that for each SP we need a separate repo from a separate build of the project?

srbarrios (approved):

terracumber_config/tf_files/Uyuni-Master-NUE.tf
```diff
@@ -117,6 +117,7 @@ module "cucumber_testsuite" {
       provider_settings = {
         mac = "aa:b2:93:01:00:d1"
       }
+      additional_repos = { java11 = "http://download.suse.de/ibs/home:/fstrba:/branches:/SUSE:/SLE-15:/Update/SUSE_SLE-15_Update/" }
```
juliogonzalez: FMPOV: @ktsamis also asking for an openSUSE 15.2 target to get a build this OS.

Yeah, I know. SLE15SP2 should be the same. But should is not the same as `is`. He can add a repo with a path to `openSUSE.org:openSUSE:Leap:15.2:Update`.

ktsamis: This one should work right? https://download.opensuse.org/repositories/Java:/Factory/openSUSE_Leap_15.2/

ktsamis: For each SP the Devel of Java/Factory has the update already, the only one that comes from the home project is 15sp3 which is not released.

ktsamis: For now we don't add it to uyuni that will come tomorrow after releasing the latest uyuni release today

PR #251, opened 2021-04-23
fix unlock

jordimassaguerpla:


jordimassaguerpla:
This fixes removing the project after the build.

PR #257, opened 2021-04-27
CI: use a copy of master project but locked

jordimassaguerpla:
This is for PR testing. In order to prevent issues when packages are
being rebuild, which happens for example when packages are in a state A
when deploying, and state B when running the tests. Some tests will try
to install packages.

Having an error because of inconsistent metadata at the test it is very
paintful because we have to retrigger a 2 hour job.

Instead, we have a project for each environment which has links to all
packages in sm:Uyuni:Master. This way, we can disable the build before
running the tests.

https://github.com/SUSE/spacewalk/issues/14654


jordimassaguerpla:
@srbarrios In order to make sure new packages are also added, I was thinking on writing a job that checks every 5 minutes if there are new packages in systemsmanagement, by using osc ls. Makes sense?

srbarrios (approved):

PR #259, opened 2021-04-28
Fix CI OBS links

jordimassaguerpla:
Links should go to sm:Uyuni:Master:TEST:XX:CR , so they get a higher
release number. Otherwise, packages could be installed from
sm:Uyuni:Master:TEST:XX:CR instead of from sm:Uyuni:Master:PR:YY


jordimassaguerpla:
@srbarrios : jfyi, I fixed this issue with obs.

PR #278, opened 2021-05-07
Use opensuse152-CI-PR for server

jordimassaguerpla:
Because it has the required dependencies, so we will go from installing
more than 900 packages, which takes 15 minutes, to 115, which takes 3.


jordimassaguerpla:
This goes after https://github.com/uyuni-project/sumaform/pull/882

PR #285, opened 2021-05-20
Fix wrong MACs after removal of centos and ubuntu, add 12sp5 buildhost in 4.1

ktsamis:
After I removed the centos 6 and ubuntu 16 I didn't fix the offset in the main.tf and now we have mismatches


Bischoff (approved):
lgtm

PR #289, opened 2021-05-26
Temporary add salt 3002 to Uyuni SLE/openSUSE

juliogonzalez:
Reason: Neither SLE15 or openSUSE Leap 15.2/3 will salt 3002 until later, and we want to start testing now.

This is not needed for other OS, because for other OS we have salt as part of the client tools, so we can adjust the links.

For SLE15/openSUSE we do not want to add salt, even temporary, as it can cause a lot of problems.


meaksh (approved):
Looks good :+1:

PR #290, opened 2021-05-27
Adapt Uyuni TFs for Leap15.3 on the server/proxy, SLE15SP3 build hosts, salt 3002 for everyone

juliogonzalez:
WARNING: This requires https://github.com/uyuni-project/sumaform/pull/890 to be merged first.

### DIFF
--- a/terracumber_config/tf_files/Uyuni-Master-NUE.tf
+++ b/terracumber_config/tf_files/Uyuni-Master-NUE.tf
@@ -92,7 +92,7 @@ module "cucumber_testsuite" {
   cc_username = var.SCC_USER
   cc_password = var.SCC_PASSWORD
 
-  images = ["centos7o", "opensuse150o", "opensuse151o", "opensuse152o", "sles15sp1o", "sles15sp2o", "ubuntu2004o"]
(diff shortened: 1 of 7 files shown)


terracumber_config/tf_files/Uyuni-Master-NUE.tf
```diff
@@ -92,7 +92,7 @@ module "cucumber_testsuite" {
   cc_username = var.SCC_USER
   cc_password = var.SCC_PASSWORD
 
-  images = ["centos7o", "opensuse150o", "opensuse151o", "opensuse152o", "sles15sp1o", "sles15sp2o", "ubuntu2004o"]
+  images = ["centos7o", "opensuse152o", "opensuse153o", "sles15sp1o", "sles15sp3o", "ubuntu2004o"]
```
juliogonzalez, line 95: I don't think we use Leap 15.0 or 15.1 for anything, anymore. The controller is 15.2 already, and the build hosts should work fine with Leap 15.3

juliogonzalez: And I don't even know why we had SLE15SP2. In the end we create the clients with SLE15SP1 :-)

juliogonzalez: Image is created now.

PR #302, opened 2021-06-15
Merge to master before running tests

jordimassaguerpla:
This way we include fixes from master branch and also new version bumps.
Otherwise, in the first case we will have failures that are already
fixed, and in the second, we will use the package from master.


jordimassaguerpla:
@srbarrios I don't know if this will work... I will try to test it with a custom job and let you know.

jordimassaguerpla:
This fixes https://github.com/SUSE/spacewalk/issues/15177

jordimassaguerpla:
> @srbarrios I don't know if this will work... I will try to test it with a custom job and let you know.

I just tested it and is working. I had to adjust the name and email of the commiter so that git can do a local merge.

PR #303, opened 2021-06-22
Use lockfile for concurrent PR e2e testing

jordimassaguerpla:
We were using test and touch for "acquiring" the test environment.

However, during our tests, we found out that two jobs could, and
actually did, take the same environment.

Most probably, this happened because two jobs run the "test" before one
of them could do the "touch".

Instead, let's use "lockfile" command, which implements semaphores,
which is what actually we were trying to implement with the "test and
touch".

Fixes https://github.com/SUSE/spacewalk/issues/15176


srbarrios (approved):
Awesome Jordi! Thanks!

PR #313, opened 2021-08-02
QA: Parametrize timeouts in BV pipelines

srbarrios:


Bischoff (approved):
LGTM

(I don't see something that changes it automatically when we are in parallel)

PR #315, opened 2021-08-04
Avoid cucumber being killed by OOM killer

Bischoff:
Avoid the test suite abrubtly stopping on the controller.

Output of dmesg:
```
[12802.008861] oom-kill:constraint=CONSTRAINT_NONE,nodemask=(null),cpuset=/,mems_allowed=0,global_oom,task_memcg=/,task=ruby.ruby2.5,pid=28148,uid=0
[12802.008925] Out of memory: Killed process 28148 (ruby.ruby2.5) total-vm:16689288kB, anon-rss:16089632kB, file-rss:0kB, shmem-rss:0kB
[12802.959889] oom_reaper: reaped process 28148 (ruby.ruby2.5), now anon-rss:0kB, file-rss:0kB, shmem-rss:0kB
```

### DIFF
--- a/terracumber_config/tf_files/SUSEManager-4.1-build-validation.tf
+++ b/terracumber_config/tf_files/SUSEManager-4.1-build-validation.tf
@@ -1220,7 +1220,7 @@ module "controller" {
   name               = "ctl"
   provider_settings = {
     mac                = "aa:b2:92:42:00:48"
-    memory             = 16384
+    memory             = 24576
     vcpu               = 8
   }
   swap_file_size = null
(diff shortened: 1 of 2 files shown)


juliogonzalez (approved):
If you have enough memory in your virtualization hosts... go for it. This would not fly for the regular testsuites :-)

Bischoff:
> If you have enough memory in your virtualization hosts... go for it. This would not fly for the regular testsuites :-)

62 GB... I'll reduce to 24 GB

The real question is: why does it use so much memory ????

Bischoff:
Was reverted.

PR #316, opened 2021-08-05
Attempt to use mortadelo on uyuni

Bischoff:
This PR:

* gets back to 4 GB for the controller on BV test suite (after we analyzed the OOMs)
* attempts to use mortadelo instead of ramrod for Uyuni master

For a start, Mortadelo will still use a storage pool `ssd` on another disk, like it is done on ramrod.
If we still see performance issues, I will change that too.


srbarrios:
@Bischoff 
4GB was the original size for BV controller?
It seems low to me.

PR #322, opened 2021-09-08
Do not call tito with verbose option

jordimassaguerpla:
Otherwise, the output is unreadable.

@srbarrios : What do you think about this one? I feel the output of the build step is unreadable and is because we are calling tito with the verbose option.


srbarrios (approved):
Might be, we don't lose nothing trying it.

PR #330, opened 2021-09-13
Java downgrade revert

ktsamis:
Use defaults again for sumaform, both repo and branch.

Also add experimental java to all branches to test it when it's available.


terracumber_config/tf_files/SUSEManager-4.1-PRV.tf
```diff
@@ -119,6 +119,9 @@ module "cucumber_testsuite" {
     server = {
       provider_settings = {
         mac = "aa:b2:92:03:00:a1"
+        additional_repos = {
+            java_TEST = "https://download.suse.de/ibs/home:/fstrba:/branches:/SUSE:/SLE-15:/Update/SUSE_SLE-15_Update/x86_64/"
```
mcalmer, line 123: Wrong URL. The last `x86_64` directory is too much. The repo starts at the place where you can find the `repodata` directory which is one level up.

ktsamis: Fixed, thanks Michael, it was a copy paste error that I didn't catch.

PR #334, opened 2021-09-16
Use our own image for opensuse clients for PR testing

jordimassaguerpla:
Otherwise, the official opensuse image keeps changing and sometimes is
not available.

Goes after https://github.com/uyuni-project/sumaform/pull/951


jordimassaguerpla:
Partial fix for https://github.com/SUSE/spacewalk/issues/15846

PR #336, opened 2021-09-17
Force a workspace name in CI jobs without @N appended

srbarrios:
Related to: https://github.com/SUSE/spacewalk/issues/15779

When our Jenkins Master is rebooted, it lost connection with our Jenkins Worker, which continue running our CI jobs.
Once it restarts, it tries to connect again witht the Jenkins worker, then the current job fails, then the next job starts but using a different workspace name, appending a `@2` there.
As we rely in terraform files that are stored in the workspace between jobs, so we don't need to re-download all the images and stuff, so we can just do a taint terraform command in the current environment... all the job fails, as the workspace is different.
This PR tries to fix this issue by using always the same workspace name.

### DIFF
--- a/jenkins_pipelines/environments/common/pipeline.groovy
+++ b/jenkins_pipelines/environments/common/pipeline.groovy
@@ -1,5 +1,6 @@
 def run(params) {
     timestamps {
+        environment_workspace = env.JOB_BASE_NAME
         // Retrieve the hash commit of the last product built in OBS/IBS and previous job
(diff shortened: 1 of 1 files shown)


jordimassaguerpla (approved):

jordimassaguerpla:
Let's try it out :)

srbarrios:
An addtional commit was added to complement this PR
 https://github.com/SUSE/susemanager-ci/commit/e5d95bf078362f7cf96caf8389ce1179610a1e84

PR #345, opened 2021-09-22
Fix client repos

jordimassaguerpla:
Use http instead of https because otherwise we get a 301 permanently
redirect error message. This is because openSUSE mirrors have issues
with https.


jordimassaguerpla:
@srbarrios This is straightforward, so I am merging. Still, please look at it fyi, it could be interesting for other issues.

PR #347, opened 2021-09-23
Add support for ARM to BV 4.2

Bischoff:
This PR adds support for ARM clients in SUSE Manager 4.2 Build Validation.

See also https://github.com/uyuni-project/sumaform/pull/957

Note: we currently use a HERE-document for the contents of the XSLT file, because `terracumber` does not coppy that file yet. This should be fixed in a subsequent PR.


srbarrios (approved):

rjmateus (approved):
lgtm

PR #349, opened 2021-09-24
Skip opensuse mirror

jordimassaguerpla:
openSUSE mirror is unreliable. For testing at the PR level, where some shortcuts are accepted, we will instead download the RPMs from obs using the API. Then, we will create a repo ourselves.

This will remove also the obs publisher from the testing. We have seen obs publisher can sometimes take a lot of time. Downloading from the build service using the API means we do not rely on the publisher.

Then, if we do this, we do not need a repo in the build service for each environment, so we can remove all the logic related to the CR subprojects. Instead, we will have a local repo in the jenkins worker for each environment.

In order to do this, we will use "osc getbinaries" for master, other and client subrepos.

For the server and proxy, we realized we also need the Other subproject, where ansible and other packages are. This is the reason why we are adding the other subproject.

We tried building the "image repo", using "osc build *.kiwi image". However, this failed locally, and autobuild couldn't give a reason. It was failing locally, while it was succeeding in the server, so I consider this unreliable as well.

One may say that getting the packages from the repo is not the same a user will have, that we should instead download them from the image repos. Yes, but this is not what we are testing at the Pull Request tests, so it is fine that we get the packages from the master and  master:other subrepo.

With this, we move the complexity from the buildservice to the worker, but all in all, this is not more complex than it was, and this should remove all those random issues about packages not being available, or not being in sync with the metadata.

At the pull request testing, we are not testing the openSUSE mirror infrastructure, so it should be fine that we skip it.

### DIFF
--- a/jenkins_pipelines/environments/common/pipeline-pull-request.groovy
+++ b/jenkins_pipelines/environments/common/pipeline-pull-request.groovy
@@ -83,35 +83,48 @@ def run(params) {
                     if(params.must_build) {
                         dir("product") {
                             // fail if packages are not building correctly
-                            sh "osc pr ${params.source_project}:TEST:${env_number}:CR -s 'F' | awk '{print}END{exit NR>1}'"
+                            sh "osc pr -r ${params.build_repo} ${params.source_project} -s 'F' | awk '{print}END{exit NR>1}'"
                             // fail if packages are unresolvable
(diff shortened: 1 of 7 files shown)


jordimassaguerpla:
This depends on https://github.com/uyuni-project/uyuni/pull/4245

jordimassaguerpla:
related to https://github.com/SUSE/spacewalk/issues/15846

srbarrios (approved):

PR #362, opened 2021-10-20
Add ARM support to 4.1 BV as well

Bischoff:
## What does this PR do?

This PR applies the 4.2 BV changes to 4.1 BV to accomodate for ARM minions.

Note: the long embedded XSLT file would be much better off as a separate file, shared between 4.1 and 4.2. But that would probably require changes to terracumber.


Bischoff:
> Support for ARM minions was not backported to 4.1. Only SLE minions are supported for the ARM architecture, not openSUSE Leap, CentOS, or CentOS Clones.

ACK. I suggest we merge this "as is" to have branches in sync, and refine afterwards.

I will add a comment though.

> We really should use a new mirror for that purpose. The mirror for x86 is almost full.

ACK. Same, let's merge it "as is" and then refine in both 4.1 and 4.2.

srbarrios:
@Bischoff sorry but reverting it for now, I need to test asap MU 4.1.12 and I can't debug the issues happening at deployment.

PR #368, opened 2021-11-10
Use terracumber from GitHub

juliogonzalez:
I just created https://github.com/uyuni-project/terracumber

A few days after the PR is merged, I will be able to remove internal repository. The few days are so the environments can run and get the new Jenkinsfiles with the new repository, and to avoid them failing.


Bischoff (approved):
lgtm

PR #369, opened 2021-11-10
correct path to terraform bin

rjmateus:
Signed-off-by: Ricardo Mateus <rmateus@suse.com>

Path to the terraform bin was wrong and point to a non existing file


juliogonzalez (approved):
@rjmateus from what I see, this worked because `terracumber` didn't care about this parameter, which is what you will fix at the other PR, right?

rjmateus:
yes that is correct. This path does exists, but terracumber don't care about this path. 
I the other PR I have changed this behavior,, and terracumber will start looking at this field.

PR #370, opened 2021-11-10
Uyuni on AWS

juliogonzalez:
- Add a new pipeline definition for Uyuni Master at AWS, as well as a new TF file
- Use the `sumaform_backend` parameter from the jobs
- Add the possibility to run jobs without tainting the environment. Mostly useful for AWS, so you don't need to pay for new instances (AWS bills by hour)

With this and https://github.com/uyuni-project/sumaform/pull/989, and I was able to launch an acceptance pipeline from a Jenkins at AWS, with no access to the R&D network at all.

Somethings are missing, however, before we can consider the work complete, but this is for other PRs, maybe even for other repositories like the testsuite.

Reference: https://github.com/SUSE/spacewalk/issues/11398

### DIFF
--- a/jenkins_pipelines/environments/common/pipeline-build-validation.groovy
+++ b/jenkins_pipelines/environments/common/pipeline-build-validation.groovy
@@ -33,7 +33,7 @@ def run(params) {
                     // Generate json file in the workspace
                     writeFile file: 'custom_repositories.json', text: params.custom_repositories, encoding: "UTF-8"
                     // Run Terracumber to deploy the environment
-                    sh "set +x; source /home/jenkins/.credentials set -x; export TF_VAR_CUCUMBER_GITREPO=${params.cucumber_gitrepo}; export TF_VAR_CUCUMBER_BRANCH=${params.cucumber_ref}; export TERRAFORM=${params.terraform_bin}; export TERRAFORM_PLUGINS=${params.terraform_bin_plugins}; ./terracumber-cli ${common_params} --logfile ${resultdirbuild}/sumaform.log ${env.TERRAFORM_INIT} --taint '.*(domain|main_disk|data_disk|server_extra_nfs_mounts).*' --custom-repositories ${WORKSPACE}/custom_repositories.json --runstep provision"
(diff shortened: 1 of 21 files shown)


jenkins_pipelines/environments/common/pipeline-build-validation.groovy
```diff
@@ -33,7 +33,7 @@ def run(params) {
...
                     // Run Terracumber to deploy the environment
-                    sh "set +x; source /home/jenkins/.credentials set -x; export TF_VAR_CUCUMBER_GITREPO=${params.cucumber_gitrepo}; export TF_VAR_CUCUMBER_BRANCH=${params.cucumber_ref}; export TERRAFORM=${params.terraform_bin}; export TERRAFORM_PLUGINS=${params.terraform_bin_plugins}; ./terracumber-cli ${common_params} --logfile ${resultdirbuild}/sumaform.log ${env.TERRAFORM_INIT} --taint '.*(domain|main_disk|data_disk|server_extra_nfs_mounts).*' --custom-repositories ${WORKSPACE}/custom_repositories.json --runstep provision"
+                    sh "set +x; source /home/jenkins/.credentials set -x; export TF_VAR_CUCUMBER_GITREPO=${params.cucumber_gitrepo}; export TF_VAR_CUCUMBER_BRANCH=${params.cucumber_ref}; export TERRAFORM=${params.terraform_bin}; export TERRAFORM_PLUGINS=${params.terraform_bin_plugins}; ./terracumber-cli ${common_params} --logfile ${resultdirbuild}/sumaform.log ${env.TERRAFORM_INIT} --taint '.*(domain|main_disk|data_disk|server_extra_nfs_mounts).*' --custom-repositories ${WORKSPACE}/custom_repositories.json --sumaform-backend ${params.sumaform_backend} --runstep provision"
```
juliogonzalez, line 36: Until now, we had the parameter on the job, but we didn't use it.

juliogonzalez:
Ping @maximenoel8 I think maybe part of the work will be useful for you, when you start deploying environments (for now my understanding is that you are still working on having the mirrors working).

Bischoff (approved):
lgtm, but i'm not the best pipelines expert

ktsamis (approved):

terracumber_config/tf_files/Uyuni-Master-AWS.tf
```diff
@@ -0,0 +1,220 @@
...
+//      provider_settings = {
+//      }
+//    }
+  }
+  provider_settings = {
+    create_network                       = false
+    public_subnet_id                     = "subnet-0f8f5847460b0be70"
+    private_subnet_id                    = "subnet-0346035ce4cd25764"
+    private_additional_subnet_id         = "subnet-0459583d3361b8aa2"
+    public_security_group_id             = "sg-0535a11916638b7e3"
```
maximenoel8, line 206: This value are set by AWS. How do you make sure you will always have this values ?

PR #372, opened 2021-11-10
Declare the commits outside the if block, so we can use it later

juliogonzalez:


Bischoff (approved):
yes, should work better

PR #373, opened 2021-11-10
use terraform bin paramter

rjmateus:
Signed-off-by: Ricardo Mateus <rmateus@suse.com>
Use terraform bin parameter


juliogonzalez (approved):
LGTM, but I am not sure if you want to adapt the other pipeline definitions at the `common` directory. Check if the environments that are using them have the parameter for terraform binary

rjmateus:
Thank for the hint. I checked and ref host is also using it.

PR #384, opened 2021-11-17
review parameters for testing PRs 

jordimassaguerpla:


srbarrios (approved):
Agree with this, we don't use that often these parameters, and it will be clear for regular use cases.

PR #390, opened 2021-11-18
Add email to the description of the job

jordimassaguerpla:
This way it is easier to identify which is yours


jordimassaguerpla:
@srbarrios jfyi. I am merging it since it is a minor thing.

PR #395, opened 2021-11-19
Adjust 'install_salt_bundle' parameter to hosts

meaksh:


juliogonzalez (commented):
Can you make install_salt_bundle = False, so it's explicit, for the Ubuntu minions?

meaksh:
The suggestion is already added at https://github.com/SUSE/susemanager-ci/commit/1bebb2f5446330fc21c508cb9448b3ebd8d628ab

PR #402, opened 2021-11-25
Two new exports, and used loops

Bischoff:
2 exports tested to work for aarch64


srbarrios (commented):
LGTM.
Maybe as refactor, do it in python using a dictionary :P

Bischoff:
> LGTM. Maybe as refactor, do it in python using a dictionary :P

I'm not sure `script()` function of HCL can start python.

PR #407, opened 2021-12-01
Acceptance tests at PR: Publish even if not build

jordimassaguerpla:
Because the repository is in a directory that depends on the
environment. If we do not publish, previous builds might be on a
different repository, so you need to publish for this environment for
the packages to be available.


srbarrios (approved):
Don't really get it :)
I just trust you

PR #417, opened 2021-12-02
Add info on lock

jordimassaguerpla:
This is to have more information on the environments. Let's store who is using it and for what PR. Also, let's put some extra info in jenkins about  it.

PR #418, opened 2021-12-02
WIP: Manage locks cleanup

jordimassaguerpla:
This goes after https://github.com/SUSE/susemanager-ci/pull/417


jordimassaguerpla:
@srbarrios This adds 2 beta parameters for keeping the build "overnight" and "over the weekend", which means 24h and 72h. If you select any of them, you will get all other builds from the same email unlocked.

Not perfect, because if they get unlocked but the job is still running, you could get the job run on that environment... so I think we should rethink this one.

jordimassaguerpla:
Closing. I don't like this idea. We could be removing locks for builds that are running and then this could be a mess with new builds.

PR #419, opened 2021-12-02
Add custom header when deploying

jordimassaguerpla:
This PR adds a custom header in the server web pages, to show that this is a test environment, who started and when to expect this to be removed. All this information is very useful because the same environment would be reused later for another user and you want to know if it is still the environment you deployed for you to debug, or if this is a new environment. This is more relevant now that we keep them for 24h, because it can be handy to know when the 24h had passed.


jordimassaguerpla:
we see now this

image

If we run "spacewalk-service restart" after the "deploy" phase. I run the command manually via ssh.

We could add that at the "remote-exec".

jordimassaguerpla:
This is how it looks 

image


image

jordimassaguerpla:
@srbarrios Can you review this one?

jordimassaguerpla:
See the text states

"In case of FAILURE you can access the environment until XXXXX"

I am using terraform to change the "custom header" to include this information. This is happening on deploy time, so I can't know if tests have failed at this point, this is why I added the text "In case of FAILURE".

I tried injecting javascript code to have a count down, but it did not worked for the login page.

jordimassaguerpla:
@srbarrios I could try instead running a command at the end of the pipeline using for example paramiko https://www.paramiko.org/

srbarrios (changes requested):
Better merge https://github.com/SUSE/susemanager-ci/pull/431

PR #429, opened 2021-12-17
Preparing BV 4.3

srbarrios:
Depends on https://gitlab.suse.de/galaxy/infrastructure/-/merge_requests/461

### DIFF
--- a/terracumber_config/tf_files/SUSEManager-4.3-build-validation.tf
+++ b/terracumber_config/tf_files/SUSEManager-4.3-build-validation.tf
@@ -1,7 +1,7 @@
 // Mandatory variables for terracumber
 variable "URL_PREFIX" {
   type = string
-  default = "https://ci.suse.de/view/Manager/view/Manager-4.2/job/manager-4.2-qa-build-validation"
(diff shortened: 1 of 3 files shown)


terracumber_config/tf_files/SUSEManager-4.3-build-validation.tf
```diff
@@ -0,0 +1,1387 @@
...
+  required_providers {
+    libvirt = {
+      source = "dmacvicar/libvirt"
+      version = "0.6.3"
+    }
+  }
+}
+
+provider "libvirt" {
+  uri = "qemu+tcp://caladan.mgr.prv.suse.net/system"
```
ktsamis, line 88: I think we need to find new servers for this BV deployment

ktsamis (approved):
Except the hypervisors that we can change later, looks great :)

terracumber_config/tf_files/SUSEManager-4.3-build-validation.tf
ktsamis: Eric will create the new servers and change these lines and then we merge

Bischoff: After discussion, merging first, and adding the servers after

PR #434, opened 2022-01-12
Use mirrored images of xen and kvm as workaround of the recurrent issue with official mirrors

srbarrios:
This PR fix the workaround made by Cédric and Jordi some days ago, so we inject the variables not from the module but directly at top of main.tf


Bischoff:
Admitting this works, this is only a workaround.

I would suggest to revert this PR, and to address the real issue, which I think is that those variables are not passed on from testsuite module to virthost module.

PR #441, opened 2022-01-21
Increase git timeout pr

maximenoel8:
Currently, git clone timeout is by default set to 10 minutes. 
To clone uyuni project, you need around 20 minutes so jenkins jobs are always failing because of this 10 minutes timeout.

This PR increase the default timeout from 10 minutes to 30 minutes.

/!\ **This PR by itself will not fix the issue**, this variable need to be added when starting the jenkins agent ( in uyuni PR case : `jenkins-worker-prs.mgr.prv.suse.net` ) : 
` -Dorg.jenkinsci.plugins.gitclient.Git.timeOut=30` 

During my tests, cloning uyuni repository in Jenkins works with this agent configuration :

`/usr/bin/java -Djava.util.logging.config.file=/usr/share/java/logging-swarm-client.properties  -Dorg.jenkinsci.plugins.gitclient.Git.timeOut=30 -jar /usr/share/java/swarm-client-jar-with-dependencies.jar -master https://ci.suse.de/ -username manager -password linux -labels pull-request-test -executors 12 -name manager-jenkins-worker-prs -fsroot /home/jenkins/jenkins-build/ -disableSslVerification -disableClientsUniqueId -mode exclusive` 

I was not able to add the variable to `manager-jenkins-worker-prs` because when I kill the process, it automatically restarts and I don't have access to Jenkins configuration.

### DIFF
--- a/jenkins_pipelines/environments/common/pipeline-pull-request.groovy
+++ b/jenkins_pipelines/environments/common/pipeline-pull-request.groovy
@@ -75,8 +75,8 @@ def run(params) {
                         //      Inside userRemoteConfigs add credentialsId: 'github'
                         checkout([  
                                     $class: 'GitSCM', 
-                                    branches: [[name: "pr/${params.pull_request_number}"]], 
-                                    extensions: [[$class: 'CloneOption', depth: 1, shallow: true]],
+                                    branches: [[name: "pr/${params.pull_request_number}"]],
+                                    extensions: [[$class: 'CloneOption', depth: 1, shallow: true, timeout: 30]],
                                     userRemoteConfigs: [[refspec: '+refs/pull/*/head:refs/remotes/origin/pr/*', url: "${pull_request_repo}"]],
                                     extensions: [
                                     [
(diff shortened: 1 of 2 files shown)


srbarrios (changes requested):
Please, let's close this PR.
We can't afford to wait 30min for something that should only take 3-4min

juliogonzalez (commented):
> Please, let's close this PR. We can't afford to wait 30min for something that should only take 3-4min

I agree this is not the proper fix: a shallow clone with depth 1 should not take that much time.

Under other circumstances, I'd say this can be merged as a workaround, but if we need to change configs in other places, then I think it's best to wait while our team is debugging this.

jordimassaguerpla:
> > Please, let's close this PR. We can't afford to wait 30min for something that should only take 3-4min
> 
> I agree this is not the proper fix: a shallow clone with depth 1 should not take that much time.
> 
> Under other circumstances, I'd say this can be merged as a workaround, but if we need to change configs in other places, then I think it's best to wait while our team is debugging this.

Closing.
Thanks @maximenoel8 in any case.

PR #446, opened 2022-01-25
Do not use openSUSE official repos

jordimassaguerpla:
We are having trouble with the official repos. From time to time there
is an error. Thus, let's not use them at all.

Use minima mirror instead of download.opensuse.org

This is to minimize depending on external infrastructure such as
openSUSE mirrors

### DIFF
--- a/terracumber_config/tf_files/Uyuni-PR-tests-env1.tf
+++ b/terracumber_config/tf_files/Uyuni-PR-tests-env1.tf
@@ -168,10 +168,17 @@ module "cucumber_testsuite" {
       provider_settings = {
         mac = "aa:b2:92:04:00:01"
       }
+      additional_repos_only = true
       additional_repos = {
         pull_request_repo = var.PULL_REQUEST_REPO,
         master_repo = var.MASTER_REPO,
         master_repo_other = var.MASTER_OTHER_REPO,
+        Update_repository_of_openSUSE_Backports = "http://minima-mirror.mgr.prv.suse.net/update/leap/15.3/backports/",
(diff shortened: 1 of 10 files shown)


jordimassaguerpla:
For this to work we need to update opensuse153-ci-pr  in the hypervisors. The image has already been updated and published in obs, we just need to make sure it is downloaded on each hypervisor.

srbarrios (approved):
LGTM.
Just be sure that those additional repos will not collide with repositories already added by sumaform in salt states. If not, all good.

jordimassaguerpla:
I am seeing an error on a package (lvm2) is not found in minima-mirror sle-updates. Seems this is not the solution.

`File './x86_64/lvm2-2.03.05-8.36.1.x86_64.rpm' not found on medium 'http://minima-mirror.mgr.prv.suse.net/update/leap/15.3/sle/'`

jordimassaguerpla:
> I am seeing an error on a package (lvm2) is not found in minima-mirror sle-updates. Seems this is not the solution.
> 
> `File './x86_64/lvm2-2.03.05-8.36.1.x86_64.rpm' not found on medium 'http://minima-mirror.mgr.prv.suse.net/update/leap/15.3/sle/'`

@moio Hi! Can you help us with this one? I don't get why the metadata is wrong in the minima-mirror.

moio:
@jordimassaguerpla I suggest you to check the repo metadata upstream, using _the same mirror minima used_ to download it.

Unfortunately, it's not the first time when we see that minima was actually doing its job well, but the upstream repo was just corrupt on the mirror.

Let me know if you discover that the upstream mirror is fine, but minima broke the repo while downloading.

PR #453, opened 2022-02-01
Add sonarqube pipeline

cbosdo:
Add a Sonarqube pipeline


jenkins_pipelines/environments/common/pipeline-sonarqube.groovy
```diff
@@ -0,0 +1,30 @@
+def run(params) {
+    stage('Clone project') {
+        git url: params.git_repo, branch: params.git_branch
+        dir ("uyuni") {
+            checkout scm
+        }
```
srbarrios, line 6: You don't need this. This will checkout susemanager-ci repository, as far as I understand you don't need it.
It's just fine with your previous line.

srbarrios:
OK, that's the groovy part that will be included in your pipeline.
Now you need your pipeline, so you pass the parameters, use as example:
https://github.com/SUSE/susemanager-ci/blob/master/jenkins_pipelines/environments/uyuni-master-infra-reference-NUE

jenkins_pipelines/environments/common/pipeline-sonarqube.groovy
cbosdo: OK

PR #473, opened 2022-02-16
Use cobbler MU for 4.1 and 4.2 CI

ktsamis:
This will be reverted after testing.


ktsamis:
this was used in jenkins, we run against the fork and branch. These MUs are tested and approved now so closing

PR #485, opened 2022-02-23
Let's not merge with git master

jordimassaguerpla:
I believe merging with master makes the git fetch/clone take longer
because a shallow copy is not enough if we want to merge the PR with
master.

This has the disadvantage that we are not including latest fixes from
master, so the developer will have to rebase her branch.

It has the advantage of being more stable and predictable.


jordimassaguerpla:
This should fix https://github.com/SUSE/spacewalk/issues/16828

srbarrios:
@jordimassaguerpla As far as I know, github does this work for us.
We have the branch from one side, and the PR from another, Github let you checkout the PR which is actually the content that mix the branch with master.

PR #486, opened 2022-02-24
QA: Removing SLES11 systems from BV4.3

srbarrios:
Removing SLES11 from our BV4.3, as we will not support it.

> --- Comment #1 from Michael Calmer <mc@suse.com> ---
> Sles11 sp4 goes EOL in a bit more than 1 month and 4.3 gets released 4 month.
> 
> SUSE Manager 4.3 will not support SLES 11.

- We are removing the retail resources too.
- We have resource tags in all Cucumber features, so when this resources are not deployed the features are skipped (including retail features)

### DIFF
--- a/terracumber_config/tf_files/SUSEManager-4.3-build-validation.tf
+++ b/terracumber_config/tf_files/SUSEManager-4.3-build-validation.tf
@@ -147,7 +147,7 @@ module "base_old_sle" {
   name_prefix = "suma-bv-43-"
   use_avahi   = false
   domain      = "mgr.prv.suse.net"
-  images      = [ "sles11sp4", "sles12sp4o", "sles12sp5o" ]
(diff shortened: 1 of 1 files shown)


srbarrios:
@mcalmer I guess we also want to remove retail support for SLES11SP4 BuildHost and SLES11SP3 terminal right?

Bischoff (approved):

Bischoff:
> @mcalmer I guess we also want to remove retail support for SLES11SP4 BuildHost and SLES11SP3 terminal right?

Rather a question for Ondrej Holecek. But he has no access to this repository apparently.

srbarrios:
@aaannz do you know what we agreed about the question above?

aaannz:
Yes, there is no support for SLE11 in 4.3 so that includes SLE11 build hosts and terminal images.

PR #487, opened 2022-02-24
do not fetch tags to save some time

jordimassaguerpla:
fix https://github.com/SUSE/spacewalk/issues/16828


jordimassaguerpla:
This adds the --no-tags option to the git fetch command. With the previous "git fetch --tags" run, it took 3 minutes to do a fetch, but with the "git fetch --no-tags" it took 45 seconds.

I expect/hope that when the "git fetch --tags" take more than 10 minutes, the "git fetch --no-tags" takes less than 10 minutes :crossed_fingers:

srbarrios (approved):

jordimassaguerpla:
@srbarrios I added the honorRefspec option , so that it only does one fetch. Without this option, we were fetching twice.

PR #489, opened 2022-02-28
Adjust Head and Test tf files to use salt bundle

meaksh:
As title mentioned, changes the TF files for the following environments to use Salt bundle (venv-salt-minion):

- head
- refhead
- testorion
- testnaica
- testhexagon

After this PR, the deployed instances will not have any "salt" package installed but instead `venv-salt-minion` package will be preinstalled. This changes also make the testsuite to know this instances are managed by `venv-salt-minion` and thefore proceed accordingly.


mcalmer (approved):
proxy is also using venv-salt-minion?

meaksh:
yes, proxy gets `venv-salt-minion` installed, so initial bootstapping happens with bundle, but then normal `salt` package is installed (but not configured), because it is required as dependency from our proxy definition for having the `/etc/salt/` folder to necessary for placing the salt broker config.

This is something we need to cleanup and readjust in the future.

PR #493, opened 2022-03-04
Use minima-mirror for server_mounted_mirror

jordimassaguerpla:
-  Use minima-mirror for SCC mirroring: stabler and faster
- this goes with https://github.com/uyuni-project/sumaform/pull/1059


jordimassaguerpla:
This makes it 1h 45' faster!!!!! I was having timeout issues with SCC and using the minima-mirror works as a charm!

PR #504, opened 2022-03-22
Check for activity in jenkins instead of total duration

ktsamis:
This is a continuation from https://github.com/SUSE/susemanager-ci/pull/502 for the rest of the testsuites. It might bring turbulence so we do it in steps.


srbarrios:
Let's close this, I tried something similar and failed due to a bug in Jenkins.
https://github.com/SUSE/susemanager-ci/pull/625

PR #510, opened 2022-03-28
add scope reportdb on test hub pipeline

mbussolotto:
According to these changes https://github.com/uyuni-project/uyuni/pull/5068 , added scope reportdb on test hub pipeline


mcalmer (approved):
As the setup is totally different, a lot of the other scopes do not make sense.
Over the long run, we should think what tests should be done on this setup.
I think this will become a new run with new scopes.
But this can be the next step and is nothing for now or this PR.

PR #525, opened 2022-04-12
Use "API" scope instead of "XMLRPC"

Bischoff:
Features tagged with `@scope_api` will use indifferently XML-RPC or HTTP APIs. This generalization needs to be reflected in the pipelines.


jordimassaguerpla (approved):

srbarrios:
I would not merge this until https://github.com/uyuni-project/uyuni/pull/5175 was merged and ported... but ok.

PR #530, opened 2022-04-26
Small fix: show pull request number on error message

jordimassaguerpla:


jordimassaguerpla:
@Bischoff Thanks for reporting this error. Here the fix. This is trivial, so I am merging it.

PR #531, opened 2022-04-27
Revert "Remove Debian 11 from 4.2"

ktsamis:
This reverts commit 265717f5ceeec4836b2d672159797556d4427184.

and we re-add debian 11


Bischoff (changes requested):
please let me redo it, we don't want to revert all of it

PR #555, opened 2022-05-20
Use pxe_boot sumaform module for Retail terminals

Bischoff:
This is only the first part on integrating better Retail into Build Validation in sumaform.

The second part will allow to set up the hardware identification from the main.tf.

### DIFF
--- a/terracumber_config/tf_files/SUSEManager-4.1-build-validation.tf
+++ b/terracumber_config/tf_files/SUSEManager-4.1-build-validation.tf
@@ -1094,21 +1094,14 @@ module "sles11sp3-terminal" {
   providers = {
     libvirt = libvirt.coruscant
   }
-  source             = "./modules/minion"
+  source             = "./modules/pxe_boot"
   base_configuration = module.base_retail.configuration
-  product_version    = "4.1-released"
   name               = "terminal-sles11sp3"
   image              = "sles11sp4" # This is not a typo
   provider_settings = {
     memory             = 1024
     vcpu               = 1
   }
-  server_configuration = {
-    hostname = "suma-bv-41-pxy.mgr.prv.suse.net"
-  }
-  auto_connect_to_master  = false
-  use_os_released_updates = false
-  ssh_key_path            = "./salt/controller/id_rsa.pub"
 }
 
 module "sles12sp5-buildhost" {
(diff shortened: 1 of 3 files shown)


terracumber_config/tf_files/SUSEManager-4.2-build-validation.tf
```diff
@@ -1224,21 +1224,14 @@ module "sles11sp3-terminal" {
   providers = {
     libvirt = libvirt.terminus
   }
-  source             = "./modules/minion"
+  source             = "./modules/pxe_boot"
   base_configuration = module.base_retail.configuration
-  product_version    = "4.2-released"
   name               = "terminal-sles11sp3"
   image              = "sles11sp4" # This is not a typo
```
vandabarata, line 1230: I know it says it's not a typo but do you know the reason why we're using sles11sp4 for a sles11sp3 terminal?

Bischoff: You build the sle 11 sp 3 terminal on a sle 11 sp 4 build host, because there is no such thing as a sle 11 sp 3 build host.

(I hope the explanation is correct - for further insights, ask Vladimir)

terracumber_config/tf_files/SUSEManager-4.1-build-validation.tf
```diff
@@ -1094,21 +1094,14 @@ module "sles11sp3-terminal" {
...
+  source             = "./modules/pxe_boot"
   base_configuration = module.base_retail.configuration
-  product_version    = "4.1-released"
   name               = "terminal-sles11sp3"
   image              = "sles11sp4" # This is not a typo
   provider_settings = {
     memory             = 1024
     vcpu               = 1
   }
-  server_configuration = {
```
vandabarata, line 1106: why do we no longer need a server configuration for the proxy the terminal connects to? did we change anything?

Bischoff: In HCL, variables are defined for each module. The `pxe_minion` module of Sumaform has less variables than the `minion` module.

And somehow it makes sense: the terminal is a perfectly innocent host that will get violently reformatted and onboarded by the proxy. Before that happens, it knows strictly nothing about any SUSE Manager server.

vandabarata (approved):

PR #564, opened 2022-06-01
Rename qa to qe

ktsamis:
Rename qa to qe in the jenkins pipelines. Together with this all these pipelines need to be configured to use the "new" files in the Script Path of the pipeline here for example https://ci.suse.de/view/Manager/view/Manager-qa/job/manager-4.1-qe-build-validation/configure


srbarrios:
> Rename qa to qe in the jenkins pipelines. Together with this all these pipelines need to be configured to use the "new" files in the Script Path of the pipeline here for example https://ci.suse.de/view/Manager/view/Manager-qa/job/manager-4.1-qe-build-validation/configure

There is one more important thing to do before this PR:
Update the current view filter in https://ci.suse.de/view/Manager/view/Manager-qa/ or other views, so we accept qe instead of qa, or both. (The jobs are taken from the ALL to a concrete view based on a regex.)

ktsamis:
I opened the ticket for this https://sd.suse.com/servicedesk/customer/portal/1/SD-88372 that you can't see and I am asking for one by one the jobs to be added. I didn't know it was from a regex, I will update the ticket

PR #570, opened 2022-06-07
do not install venv-salt-minion in sles15sp3 instance

mbussolotto:
do not install venv-salt-minion in sles15sp3 instance


Bischoff (commented):
I don't know whether those 2 lines are still needed or if we can remove them already.

Maybe Julio knows. Or someone from ion.

mcalmer:
This should only be needed until the venv-salt-minion is released. But ok for now

mcalmer (approved):

PR #543, opened 2022-05-11
QE: Switch HEAD CI to use SLES15SP4

nodeg:
**DO NOT MERGE YET!**

After discussion with the rest of the QE team, we want to temporarily switch our HEAD CI to use a custom branch prepared for using SLES15SP4. ATM we only want to have this branch with the changes not present in master.

Issue: https://github.com/SUSE/spacewalk/issues/17410
PR: https://github.com/SUSE/spacewalk/pull/17629

PR #594, opened 2022-07-04
QE: Switch Uyuni CI to use SLES15SP4 and Leap 15.4

nodeg:
We move the testsuite CI to use SLES15SP4. We made the changes only on Uyuni (and therefore HEAD) so we have to adjust the Uyuni `.tf`, too otherwise the testsuite will not work correctly.

See:
- https://github.com/SUSE/spacewalk/issues/17410
- https://github.com/uyuni-project/uyuni/pull/5413
- https://github.com/SUSE/susemanager-ci/pull/543

### DIFF
--- a/terracumber_config/tf_files/Uyuni-Master-NUE.tf
+++ b/terracumber_config/tf_files/Uyuni-Master-NUE.tf
@@ -151,6 +151,7 @@ module "cucumber_testsuite" {
       install_salt_bundle = true
     }
     suse-minion = {
+      # left with SP3 since we update it to SP4 in the testsuite
       image = "sles15sp3o"
       name = "min-sles15"
       provider_settings = {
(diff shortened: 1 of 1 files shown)


ktsamis (approved):

rjmateus (approved):
LGTM

terracumber_config/tf_files/Uyuni-Master-NUE.tf
```diff
@@ -151,7 +151,7 @@ module "cucumber_testsuite" {
       install_salt_bundle = true
     }
     suse-minion = {
-      image = "sles15sp2o"
+      image = "sles15sp3o"
```
srbarrios, line 154: Let's do something, because this is confusing.
Let's add a comment and explain that we use a previous SP, so we upgrade it inside the test suite. Make sense?

nodeg: Good idea.

nodeg: Added.

srbarrios (approved):

PR #607, opened 2022-07-15
Add `terraform apply -parallelism` parameter

nodeg:
This will add the `terraform apply -parallelism` parameter to (hopefully) speed up the deployment stage. I increased the value from the default 10 to 20. When this works, maybe we can increase it even further on e.g. one pipeline to test it there.

### Before merging this
- [x] Get the go from Julio on the unit tests for terracumber
- [x] merge https://github.com/uyuni-project/terracumber/pull/11

### Links
- https://github.com/SUSE/spacewalk/issues/18101
- https://github.com/uyuni-project/terracumber/pull/11

### DIFF
--- a/jenkins_pipelines/environments/common/pipeline-build-validation.groovy
+++ b/jenkins_pipelines/environments/common/pipeline-build-validation.groovy
@@ -6,7 +6,7 @@ def run(params) {
         // The junit plugin doesn't affect full paths
         junit_resultdir = "results/${BUILD_NUMBER}/results_junit"
 
-        if (!params.terraform_parallelism) {
+        if (!params.terraform_parallelism || params.terraform_parallelism == '') {
             params.terraform_parallelism = 10
         }
 
--- a/jenkins_pipelines/environments/common/pipeline-build-validation.groovy
+++ b/jenkins_pipelines/environments/common/pipeline-build-validation.groovy
@@ -5,7 +5,7 @@ def run(params) {
         env.resultdirbuild = "${resultdir}/${BUILD_NUMBER}"
         // The junit plugin doesn't affect full paths
         junit_resultdir = "results/${BUILD_NUMBER}/results_junit"
-        env.common_params = "--outputdir ${resultdir} --tf ${params.tf_file} --gitfolder ${resultdir}/sumaform"
+        env.common_params = "--outputdir ${resultdir} --tf ${params.tf_file} --gitfolder ${resultdir}/sumaform --parallelism ${params.terraform_parallelism}"
         try {
             stage('Clone terracumber, susemanager-ci and sumaform') {
                 // Create a directory for  to place the directory with the build results (if it does not exist)
--- a/jenkins_pipelines/environments/common/pipeline-build-validation.groovy
+++ b/jenkins_pipelines/environments/common/pipeline-build-validation.groovy
@@ -5,6 +5,11 @@ def run(params) {
         env.resultdirbuild = "${resultdir}/${BUILD_NUMBER}"
         // The junit plugin doesn't affect full paths
         junit_resultdir = "results/${BUILD_NUMBER}/results_junit"
+
+        if (!params.terraform_parallelism) {
+            params.terraform_parallelism = 10
+        }
+
         env.common_params = "--outputdir ${resultdir} --tf ${params.tf_file} --gitfolder ${resultdir}/sumaform --parallelism ${params.terraform_parallelism}"
         try {
             stage('Clone terracumber, susemanager-ci and sumaform') {
--- a/jenkins_pipelines/environments/common/pipeline-build-validation.groovy
+++ b/jenkins_pipelines/environments/common/pipeline-build-validation.groovy
@@ -6,11 +6,11 @@ def run(params) {
         // The junit plugin doesn't affect full paths
         junit_resultdir = "results/${BUILD_NUMBER}/results_junit"
 
-        if (!params.terraform_parallelism || params.terraform_parallelism == '') {
-            params.terraform_parallelism = 10
-        }
+        env.common_params = "--outputdir ${resultdir} --tf ${params.tf_file} --gitfolder ${resultdir}/sumaform"
 
-        env.common_params = "--outputdir ${resultdir} --tf ${params.tf_file} --gitfolder ${resultdir}/sumaform --parallelism ${params.terraform_parallelism}"
+        if (params.terraform_parallelism) {
+            env.common_params = "${env.common_params} --parallelism ${params.terraform_parallelism}"
+        }
         try {
             stage('Clone terracumber, susemanager-ci and sumaform') {
                 // Create a directory for  to place the directory with the build results (if it does not exist)
--- a/jenkins_pipelines/environments/common/pipeline-pull-request.groovy
+++ b/jenkins_pipelines/environments/common/pipeline-pull-request.groovy
@@ -208,7 +208,7 @@ def run(params) {
                         env.resultdirbuild = "${resultdir}/${BUILD_NUMBER}"
                         env.tf_file = "susemanager-ci/terracumber_config/tf_files/Uyuni-PR-tests-env${env_number}.tf" //TODO: Make it possible to use environments for SUMA
 
-                        if (!params.terraform_parallelism) {
+                        if (!params.terraform_parallelism || params.terraform_parallelism == '') {
                             params.terraform_parallelism = 10
                         }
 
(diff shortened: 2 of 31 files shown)


jenkins_pipelines/environments/manager-4.3-dev-acceptance-tests-NUE
```diff
@@ -14,6 +14,7 @@ node('sumaform-cucumber') {
             choice(name: 'sumaform_backend', choices: ['libvirt', 'aws'], description: 'Sumaform backend to be used (see https://github.com/uyuni-project/sumaform#backend-choice)'),
             choice(name: 'terraform_bin', choices: ['/usr/bin/terraform'], description: 'Terraform binary path'),
             choice(name: 'terraform_bin_plugins', choices: ['/usr/bin'], description: 'Terraform plugins path'),
+            choice(name: 'terraform_parallelism', choices: ['20'], description: 'Define the number of parallel resource operations for terraform'),
```
srbarrios, line 17: Instead of using an unique choice, that will not let as parametrize this value when we run the pipeline... better use 20 as default value of a string.

```suggestion
             string(name: 'terraform_parallelism', defaultValue: '20', description: 'Define the number of parallel resource operations for terraform'),
```

srbarrios (approved):
I would only like to change the Jenkins parameter, instead of using a choice parameter, better use a string parameter with a default value.

jordimassaguerpla:
https://github.com/SUSE/susemanager-ci/blob/master/jenkins_pipelines/environments/uyuni-prs-ci-tests

nodeg:
Of course @jordimassaguerpla.

jenkins_pipelines/environments/common/pipeline-reference.groovy
```diff
@@ -3,6 +3,11 @@ def run(params) {
         deployed = false
         env.resultdir = "${WORKSPACE}/results"
         env.resultdirbuild = "${resultdir}/${BUILD_NUMBER}"
+
+        if (!params.terraform_parallelism) {
```
srbarrios, line 7: I'm not sure if in groovy that's the way to handle an element not defined.
If that's ok, then good:)

jenkins_pipelines/environments/common/pipeline-build-validation.groovy
```diff
@@ -5,7 +5,12 @@ def run(params) {
         env.resultdirbuild = "${resultdir}/${BUILD_NUMBER}"
         // The junit plugin doesn't affect full paths
         junit_resultdir = "results/${BUILD_NUMBER}/results_junit"
-        env.common_params = "--outputdir ${resultdir} --tf ${params.tf_file} --gitfolder ${resultdir}/sumaform"
+
+        if (!params.terraform_parallelism) {
+            params.terraform_parallelism = 10
+        }
+
+        env.common_params = "--outputdir ${resultdir} --tf ${params.tf_file} --gitfolder ${resultdir}/sumaform --parallelism ${params.terraform_parallelism}"
```
srbarrios, line 13: If I may suggest...
I would design it differently.

I think the default value must be managed and coming from Terracumber:
https://github.com/uyuni-project/terracumber/pull/11/files#diff-56646f4f9cb011c599f92d8cd92395d82e90023b111c80931c0dcf8d8267bf47R51-R53

And so, better not handle the default value two times.
So, what I would do is something like... (not tested):

```suggestion
        env.common_params = "--outputdir ${resultdir} --tf ${params.tf_file} --gitfolder ${resultdir}/sumaform"

        if (params.terraform_parallelism) {
            env.common_params = "${env.common_params} --parallelism ${params.terraform_parallelism}"
        }
```

(Keep it only as example, don't directly commit the suggestion :D )

nodeg: Yes, the default is coming from terracumber, so I think we can skip the whole check for the parameter so it should always be set there. Either with the default value (10) or the value set by the user.
What do you say @jordimassaguerpla?

jenkins_pipelines/environments/common/pipeline-pull-request.groovy
```diff
@@ -206,7 +206,12 @@ def run(params) {
                         env.resultdir = "${WORKSPACE}/results"
                         env.resultdirbuild = "${resultdir}/${BUILD_NUMBER}"
                         env.tf_file = "susemanager-ci/terracumber_config/tf_files/Uyuni-PR-tests-env${env_number}.tf" //TODO: Make it possible to use environments for SUMA
-                        env.common_params = "--outputdir ${resultdir} --tf ${tf_file} --gitfolder ${resultdir}/sumaform"
+
+                        if (!params.terraform_parallelism) {
```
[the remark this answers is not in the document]
nodeg: What I already mentioned above in Oscar's suggestion. I think there is no possibility that this variable is empty or not set since we always set a value in terracumber. See comment above.

jenkins_pipelines/environments/uyuni-prs-ci-tests
```diff
@@ -14,6 +14,7 @@ node('pull-request-test') {
             string(name: 'cucumber_gitrepo', defaultValue: 'https://github.com/uyuni-project/uyuni.git', description: 'Advanced: Change this by your repo, only if you changed the tests in your PR'),
             string(name: 'cucumber_ref', defaultValue: 'master', description: 'Advanced: Change this by your branch, only if you changed the tests in your PR'),
             string(name: 'additional_repo_url', defaultValue: '', description: 'Advanced: Add the URL of an additional repo to test new packages or package updates (only for server, proxy, kvm and xen hosts).'),
+            string(name: 'terraform_parallelism', defaultValue: '20', description: 'Define the number of parallel resource operations for terraform'),
```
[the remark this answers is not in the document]
nodeg: fixed.

jenkins_pipelines/environments/common/pipeline-reference.groovy
nodeg: Jep, groovy is able to handle this. I looked at some other examples. Jordi also commented below.

nodeg:
Rebase done.

jenkins_pipelines/environments/common/pipeline-build-validation.groovy
jordimassaguerpla: I would keep the check just in case. We could change the parameters list later on and forget about this. It is to be on the safe side.

jenkins_pipelines/environments/common/pipeline-pull-request.groovy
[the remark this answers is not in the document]
jordimassaguerpla: I would just keep it to be on the safe side. That variable is coming from another file, so a developer could change it later on.

jenkins_pipelines/environments/common/pipeline-build-validation.groovy
jordimassaguerpla: I like Oscar's suggestion.

jordimassaguerpla (commented):
I added some comments. Generally, this lgtm, so no need for more revisions from my side. I would say, let's give it a try :)

nodeg:
Alright. Thanks Jordi. Then we only need to wait for the go from Julio on the terracumber front and I would appreciate if you would approve this PR the usual way 😉

jordimassaguerpla (approved):

jenkins_pipelines/environments/common/pipeline-build-validation.groovy
srbarrios: Not sure if you get the detail of `env.common_params = "${env.common_params} --parallelism ${params.terraform_parallelism}"` the idea is that if the variable is defined, we add the parameter and its value, otherwise we don't add anything and therefore it will pick the default value from terracumber.

That way we don't handle a default value two times, one in Terracumber and another in the Jenkins Pipeline.

jenkins_pipelines/environments/common/pipeline-build-validation.groovy
```diff
@@ -5,7 +5,12 @@ def run(params) {
         env.resultdirbuild = "${resultdir}/${BUILD_NUMBER}"
         // The junit plugin doesn't affect full paths
         junit_resultdir = "results/${BUILD_NUMBER}/results_junit"
-        env.common_params = "--outputdir ${resultdir} --tf ${params.tf_file} --gitfolder ${resultdir}/sumaform"
+
+        if (!params.terraform_parallelism || params.terraform_parallelism == '') {
+            params.terraform_parallelism = 10
```
srbarrios, line 10: Either we have a default value here, either we have a default value in Terracumber, both... I don't see it.

nodeg: Ok, then I would say we remove it here as use the default value from terracumber.

PR #620, opened 2022-07-28
QE: Recover timeout by unactivity in our test pipelines

srbarrios:
In the past our pipelines were using activity timeout, instead of just a total timeout.
I think we should recover this approach, so we don't depend on how the test takes in time, but if it's doing something or not.

### DIFF
--- a/jenkins_pipelines/environments/manager-4.1-dev-acceptance-tests-NUE
+++ b/jenkins_pipelines/environments/manager-4.1-dev-acceptance-tests-NUE
@@ -29,7 +29,7 @@ node('sumaform-cucumber') {
     stage('Checkout pipeline') {
         checkout scm
     }
-    timeout(activity: false, time: 10, unit: 'HOURS') {
+    timeout(activity: true, time: 3, unit: 'HOURS') {
         def pipeline = load "jenkins_pipelines/environments/common/pipeline.groovy"
         pipeline.run(params)
     }
--- a/jenkins_pipelines/environments/manager-4.1-dev-acceptance-tests-PRV
+++ b/jenkins_pipelines/environments/manager-4.1-dev-acceptance-tests-PRV
@@ -29,7 +29,7 @@ node('sumaform-cucumber') {
     stage('Checkout pipeline') {
         checkout scm
     }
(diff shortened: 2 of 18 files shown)


juliogonzalez (approved):
LGTM. I see some environments such as `manager-4.1-dev-acceptance-tests-NUE` are not part of this PR, but I guess they still have the old timeout?

jenkins_pipelines/environments/manager-mu-aws
```diff
@@ -28,7 +28,7 @@ node('sumaform-cucumber') {
     stage('Checkout pipeline') {
         checkout scm
     }
-    timeout(activity: false, time: 10, unit: 'HOURS') {
+    timeout(activity: false, time: 20, unit: 'HOURS') {
```
nodeg, line 31:
```suggestion
    timeout(activity: false, time: 2, unit: 'HOURS') {
```

jenkins_pipelines/environments/uyuni-mu-cloud
```diff
@@ -28,7 +28,7 @@ node('sumaform-cucumber') {
     stage('Checkout pipeline') {
         checkout scm
     }
-    timeout(activity: false, time: 10, unit: 'HOURS') {
+    timeout(activity: false, time: 20, unit: 'HOURS') {
```
nodeg, line 31:
```suggestion
    timeout(activity: false, time: 2, unit: 'HOURS') {
```

nodeg (commented):
You double the timeout for clouds. Was this intentional?

jenkins_pipelines/environments/manager-mu-aws
srbarrios: We don't want to use activity timeout in cloud, as we don't want in PR tests.
Because we have some activities like mirroring content, or building content, that doesn't print feedback in the console output.

jenkins_pipelines/environments/uyuni-mu-cloud
srbarrios: Same than above

nodeg (approved):
LGTM. Let's see how this affects our current situation and if we see more timeouts or not.

elariekerboull (approved):

PR #633, opened 2022-08-12
Use a different label for jenkins worker in Provo

Bischoff:
Use a different label for jenkins worker in Provo

The new label is defined in the infrastructure pillars


srbarrios (approved):
https://ci.suse.de/label/sumaform-cucumber-provo/
LGTM

PR #636, opened 2022-08-17
Add Retail to BV Jenkins pipeline for 4.3/4.2

nodeg:
Fixes https://github.com/SUSE/spacewalk/issues/17973

### DIFF
--- a/jenkins_pipelines/environments/common/pipeline-build-validation.groovy
+++ b/jenkins_pipelines/environments/common/pipeline-build-validation.groovy
@@ -118,6 +118,20 @@ def run(params) {
                 }
             }
 
+            stage('Prepare and run Retail') {
+                if(params.must_prepare_retail) {
+                    echo 'Prepare Proxy for Retail'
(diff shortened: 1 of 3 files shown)


vandabarata (approved):

Bischoff (approved):
it looks good, but I am no real expert for pipelines :-( 

we must probably just test and fix at the end of 4.3.1

jordimassaguerpla (approved):
I am not familiar with this pipeline specifically ... still I have not spotted any "weird" thing, so I agree with Eric, merge and try, and fix if needed afterwards.

jordimassaguerpla:
On a second thought ... you are adding a new feature that was not there. You could add that and by default have it disabled, so you can try it without breaking the actual workflow, and then, once you are sure this works, you can make it enabled by default. Just a suggestion.

jordimassaguerpla:
like a feature flag on trunk based development https://trunkbaseddevelopment.com/feature-flags/ ... so you merge the new code in master but disabled, and when you are sure, you only have to enable it.

Bischoff:
> On a second thought ... you are adding a new feature that was not there. You could add that and by default have it disabled, so you can try it without breaking the actual workflow, and then, once you are sure this works, you can make it enabled by default. Just a suggestion.

the actual workflow is only ran by hand....

so it being "disabled" does not change much

jordimassaguerpla:
> > On a second thought ... you are adding a new feature that was not there. You could add that and by default have it disabled, so you can try it without breaking the actual workflow, and then, once you are sure this works, you can make it enabled by default. Just a suggestion.
> 
> the current workflow is only ran by hand in this case...
> 
> so it being "disabled" does not change much

ok. Then go ahead merging :)

nodeg:
Thanks Jordi!

PR #646, opened 2022-08-31
Add Ubuntu 22.04 to 4.3 BV

nodeg:
This will add Ubuntu 22.04 to the 4.3 BV.

### DIFF
--- a/terracumber_config/tf_files/SUSEManager-4.3-build-validation.tf
+++ b/terracumber_config/tf_files/SUSEManager-4.3-build-validation.tf
@@ -248,7 +248,7 @@ module "base_debian" {
   name_prefix = "suma-bv-43-"
   use_avahi   = false
   domain      = "mgr.prv.suse.net"
-  images      = [ "ubuntu1804o", "ubuntu2004o", "debian10o", "debian11o" ]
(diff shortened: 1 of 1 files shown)


Bischoff (changes requested):
you forgot the ssh minion ?

terracumber_config/tf_files/SUSEManager-4.3-build-validation.tf
```diff
@@ -815,29 +815,29 @@ module "ubuntu2004-minion" {
...
+    memory             = 4096
+  }
+  server_configuration = {
+    hostname = "suma-bv-43-pxy.mgr.prv.suse.net"
+  }
+  auto_connect_to_master  = false
+  use_os_released_updates = false
+  ssh_key_path            = "./salt/controller/id_rsa.pub"
+
+  //ubuntu2204-minion_additional_repos
```
srbarrios, line 838: Context:
We had this comment in the past with the idea to support in Terracumber additional repositories for MUs directly injected during deployment.
Nowadays, we decided to do it properly, so we create custom channels inside SUMA server, and add the custom channel to the AK used during the bootstrap of the under register system.
So, I think... this comment become unuseful.

nodeg: Thank you for providing some context on this.

Bischoff (approved):

PR #655, opened 2022-09-21
Output project url in case of build failure

jordimassaguerpla:


Etheryte (approved):
This is a great improvement 👍

PR #659, opened 2022-10-03
Change MAC Address to correspond them to Ubuntu 22

srbarrios:


srbarrios:
We will need to add and replace some MAC addresses and entries in our dhcp too.

PR #676, opened 2022-10-19
Specify image and names in configuration files

nodeg:
This will specify an image and a name for each node in our terraform configuration files.

Fixes https://github.com/SUSE/spacewalk/issues/19270

### DIFF
--- a/terracumber_config/tf_files/SUSEManager-4.1-NUE.tf
+++ b/terracumber_config/tf_files/SUSEManager-4.1-NUE.tf
@@ -173,6 +173,7 @@ module "cucumber_testsuite" {
     }
     build-host = {
       image = "sles15sp2o"
+      name = "min-build"
       provider_settings = {
         mac = "aa:b2:93:01:00:ad"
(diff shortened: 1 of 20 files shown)


jordimassaguerpla (approved):

terracumber_config/tf_files/SUSEManager-4.1-NUE.tf
```diff
@@ -173,6 +173,7 @@ module "cucumber_testsuite" {
     }
     build-host = {
       image = "sles15sp2o"
+      name = "min-build"
```
srbarrios, line 176: build-host, kvm-host, xen-host might be also not necessary...

srbarrios (approved):

terracumber_config/tf_files/SUSEManager-4.1-NUE.tf
nodeg: Yes but now I already did the changes 😉

PR #685, opened 2022-11-02
Use minima-mirror for 15.4 source

elariekerboull:
Switch the mirror source from `minima-mirror-bv` to `minima-mirror` for 15.4 product on 4.2 BV tf file.


ktsamis (approved):
looks great, thanks for finding and fixing this

PR #696, opened 2022-11-17
Switch from metropolis to hugo

ktsamis:
Switch from metropolis to hugo because metropolis is down and it might need some time to be up again.


Bischoff (approved):
thanks lgtm

mcalmer:
uhhh, don't you have something faster? :-)
Hugo is one of the oldest beast we have

PR #708, opened 2022-11-29
Install salt bundle on 4.3 BV

Bischoff:
In the very early stages of preparation of the VMs, sumaform needs a Salt client to apply its configuration states. In the case of "official" images, `salt-minion` might be on the image, otherwise we install it from the `cloud-init`. We can either install `salt-minion` or `venv-salt-minion`, depending on the main.tf.

In 4.3 (but not 4.2) we need to pick the venv version. Especially that on some recent minions the `salt-minion` package might not even be available anymore in the tools. This PR addresses this.

Note: this PR sets this flag also on the SSH minions, because sumaform uses Salt to set up the SSH minions too.


srbarrios (approved):
It looks what we agreed, so LGTM.

ktsamis (approved):

meaksh (approved):

PR #710, opened 2022-11-30
Revert: don't install the salt bundle on 4.3 BV

Bischoff:
Revert of #708. After detailed discussion with Pablo, it appears the original PR was at best useless:

* sumaform always uses salt-minion (no venv) to do its initial stuff
* BV test suite does its own stuff (first thing it does is remove all Salt stuff coming from sumaform)

The problem of absent salt-minion package on Rocky 9 might be an effect of https://bugs.rockylinux.org/view.php?id=1189 . If not, the reverted PR would not have solved the issue as sumaform always uses salt-minion, and we have to find something else.


srbarrios (approved):
Make sense. After reading the description. Ok.

meaksh (approved):

PR #713, opened 2022-12-07
Raising available memory for SUMA server to 10GB on 4.2 CI

elariekerboull:
Following https://github.com/SUSE/spacewalk/issues/19666, raising the available memory on the 4.2 SUSE Manager CI server from default 8GB to 10GB.

4.3 raised to 10GB as well
Head raised to 12GB to mirror Uyuni config

PR #714, opened 2022-12-09
Revert memory raise for suma CIs

elariekerboull:
Since raising the memory to 10GB did not bring an expected fix to some issues on 4.2 , reverting to default memory attribution on 4.2 and 4.3 (8GB) to avoid raising it for no reason.
Original PR: https://github.com/SUSE/susemanager-ci/pull/713

Related card: https://github.com/SUSE/spacewalk/issues/19666


Bischoff (approved):
ok but beware we also have 4.2-NUE.tf and 4.3-NUE.tf

PR #720, opened 2022-12-14
Availability zone in Uyuni AWS changed to eu-central-1c

srbarrios:
This PR aims to prevent that kind of issues in AWS due to not available resources in the current availability zone.
```
14:25:21  │ Error: Error launching source instance: InsufficientInstanceCapacity: We currently do not have sufficient c6a.xlarge capacity in the Availability Zone you requested (eu-central-1a). Our system will be working on provisioning additional capacity. You can currently get c6a.xlarge capacity by not specifying an Availability Zone in your request or choosing eu-central-1b, eu-central-1c.
```

There is three uncertain possible issues:
- What happen with old instances? They are clean up?
- If that change implies a different subnet, will sumaform handle this change properly so we can continue using the mirror in the other subnet?
- What happen with the hardcoded resource for our mirror, will that cause a not desired kind of traffic?

### DIFF
--- a/terracumber_config/tf_files/Uyuni-Master-AWS.tf
+++ b/terracumber_config/tf_files/Uyuni-Master-AWS.tf
@@ -81,7 +81,7 @@ variable "REGION" {
 
 variable "AVAILABILITY_ZONE" {
   type = string
-  default = "eu-central-1a"
+  default = "eu-central-1c"
 }
 
 variable "KEY_FILE" {


juliogonzalez (approved):
LGTM, but @rjmateus should confirm what will happen with the old resources.

juliogonzalez:
> LGTM, but @rjmateus should confirm what will happen with the old resources.

And you will need to adjust the subnet as well. Not sure if we have a subnet for this new AZ, created by the separate tf that handles the basic infra + the bastion host.

@rjmateus designed and developed it, so he should know.

mcalmer (changes requested):
It is not that easy. As julio said, the infrastructure is missing in this zone.
Maybe we can use a different instance type if this one is not available.

mcalmer:
c6a - the "a" is for AMD. AFAIK we can try "c6i" (i == intel) or we try something different.

rjmateus:
@juliogonzalez currently @mcalmer is implementing a PR on sumaform which allow is to just create the network. I would say sumaform should not create networks on all AZ. But with this PR from Michael we will have the flexibility to create extra networks in other AZ's if needed, re-using the same VPC

juliogonzalez:
> @juliogonzalez currently @mcalmer is implementing a PR on sumaform which allow is to just create the network. I would say sumaform should not create networks on all AZ. But with this PR from Michael we will have the flexibility to create extra networks in other AZ's if needed, re-using the same VPC

What PR from Michael?

But anyway, I don't see why not: subnets are free :-)

And it's not just about creating subnets, but about configuring security groups as well, as subnets are mentioned there... unless we are using internal IP ranges at the security groups for some strange reason.

srbarrios:
Closing, I let @mcalmer handle this issue with his PR.

rjmateus:
@juliogonzalez I'm talking about this one: https://github.com/uyuni-project/sumaform/pull/1220
We limit the access of subnets with security groups IP Ranges

PR #722, opened 2022-12-15
Fix Orion environment

mbussolotto:
Align `SUSEManager-Test-Orion-NUE.tf` to `SUSEManager-Test-NUE.tf`


mackdk (approved):

mcalmer (approved):
give it a try. Not sure if head is already usable. But we are on the way.

PR #725, opened 2022-12-21
Adding a confirmation step before each stage in BV

srbarrios:
Most of us when running the BV, prefer to don't run it all at once, but run only some stages and verify all is good before moving to next stage. Or even, made some trick in between.

In order to facilitate having all in a unique job execution, I added confirmation messages that will wait until we type a key before running the next stage.


vandabarata (approved):

nodeg (approved):
LGTM although I am not sure if @Bischoff is a big fan of that change.

srbarrios:
> LGTM although I am not sure if @Bischoff is a big fan of that change.

Actually I'm doing it to make him a but happier. :)

PR #727, opened 2022-12-27
Remove sle15 from BVs

ktsamis:
Remove sle15 from BVs


vandabarata (approved):

ktsamis:
I think this is safe to merge, we won't deploy soon

PR #737, opened 2023-01-13
Extend test suite information in the README

nodeg:
This will add some of the frequent information when talking about the test suite like on what hypervisor it runs and if it uses a mirror.

### DIFF
--- a/README.md
+++ b/README.md
@@ -53,7 +53,9 @@ the build is reproducible.
 image for testing Pull Requests built with the open build service. This needs to be copied manually.
 - http://minima-mirror.mgr.prv.suse.net/distribution/leap/: oss and non-oss repositories.
 
-## Used image versions in the CI test suite
+## CI Test suite
+
+### Used image versions
 
 | Version | Minion      | SSH minion  | Client      | RH-like  | Deb-like     | Virthosts   | Buildhost   | Terminal    | Controller | Server      | Proxy       |
 | ------- | ----------- | ----------- | ----------- | -------- | ------------ | ----------- | ----------- | ----------- | ---------- | ----------- | ----------- |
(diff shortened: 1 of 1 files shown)


elariekerboull (approved):
Great idea, thanks for that!

README.md
```diff
@@ -53,7 +53,9 @@ the build is reproducible.
 image for testing Pull Requests built with the open build service. This needs to be copied manually.
 - http://minima-mirror.mgr.prv.suse.net/distribution/leap/: oss and non-oss repositories.
 
-## Used image versions in the CI test suite
+## CI Test suite
```
vandabarata, line 56:
```suggestion
## CI Testsuite
```

vandabarata (approved):

README.md
nodeg: Nope. We use `test suite` for quite some time now 😉

srbarrios (approved):

Bischoff (commented):
Ah, that is a recurring problem.

We said several times that this information changes too often to be put in some documentation file.
The current theory is to look in grafana - however that is not convenient.

In short, I like this PR, but it goes against what we said previously. That's not to say we cannot change our mind.

Small remark: we could add information for BV (but then have to change the title).

srbarrios (approved):
I would say it's enough maintaining this info in the infrastructure repository and also in our grafana dashboards. This will cause some extra maintenance. But I would not oppose to merge it, just giving my personal opinion about that kind of doc.

nodeg:
I am also not a big fan of having duplicate information but sometimes I need information and I do not know where to find it. We could also of course extend our Confluence QE landing page with all the information or links to the information to have everything in in one place. So we instantly know where to look.

nodeg:
Ok. I guess I am closing this PR in favor of adding/linking the information in our Confluence space. Anyway thank you for your reviews and comments.

PR #742, opened 2023-01-20
Increase disk size in 4.3 BV server

ktsamis:
Increase disk size in 4.3 BV server because we are at the limit:

```
/dev/vdb1                                      1.5T  1.3T  119G  92% /var/spacewalk
```

### DIFF
--- a/terracumber_config/tf_files/SUSEManager-4.3-build-validation.tf
+++ b/terracumber_config/tf_files/SUSEManager-4.3-build-validation.tf
@@ -299,7 +299,7 @@ module "server" {
   }
 
   server_mounted_mirror = "minima-mirror-bv.mgr.prv.suse.net"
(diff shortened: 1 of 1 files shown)


Bischoff (approved):

terracumber_config/tf_files/SUSEManager-4.3-build-validation.tf
```diff
@@ -299,7 +299,7 @@ module "server" {
   }
 
   server_mounted_mirror = "minima-mirror-bv.mgr.prv.suse.net"
-  repository_disk_size = 1500
+  repository_disk_size = 2000
```
srbarrios, line 302:
```suggestion
  repository_disk_size = 1700
```

Let's be more conservative.

PR #743, opened 2023-01-22
[AWS] AWS BV 4.3 

maximenoel8:
New CI pipeline to support BV testing on AWS. This PR add SUMA 4.3 configuration.

Two mirrors are deployed during this pipeline. They are meant to manage MU internal repositories use in public cloud AWS.


terracumber_config/tf_files/local_mirror.tf
```diff
@@ -122,11 +123,16 @@ module "mirror" {
...
+  image = "opensuse154o"
   volume_provider_settings = {
     pool = local.pool
     // uncomment next line to use existing snapshot as starting point
     //    volume_snapshot_id = data.aws_ebs_snapshot.data_disk_snapshot.id
   }
+  provider_settings = {
+    mac                = "52:54:00:ba:b7:98"
+    memory             = 4096
+  }
```
Bischoff, line 135:
```
grog# virsh list
(...)
 139   susemanager-4-3-aws-build-validation-mirror   en cours d’exécution
```
--> nice, but will that be set up automatically next time? I see no `name =` directive in this maint.tf file.

Bischoff (commented):
Very nice, just a question left

mcalmer (approved):

srbarrios:
@maximenoel8 I suggest you merge this PR as soon as possible and don't add more commits on top of this one.
We will not be able to run this pipeline anytime soon in any case. As BV in AWS has as a blocker the need to identify which IBS MI repos have or not CVS before we push it to cloud.

I think it will be good we work in a new card, to automatically identify CVS in a set of submissions and stop the workflow in a early stage (probably pushing content to our mirror).

PR #769, opened 2023-02-16
Temporarily disable SLE micro in BV

ktsamis:
Temporarily disable SLE micro in BV because it fails to install timezone package because of the read-only filesystem


nodeg (approved):
LGTM but I wonder if `//` will work as comment. I always use `#` for commenting.

ktsamis:
> LGTM but I wonder if `//` will work as comment. I always use `#` for commenting.

it's for HCL an alternate way to comment, I'm used to it now

PR #772, opened 2023-02-17
QE: Add new variables for nested VM tests

nodeg:
This adds a new variable to our terraform configuration files necessary for the Salt bundle migration tests that uses nested VMs.

See
- https://github.com/SUSE/spacewalk/issues/17238
- https://github.com/uyuni-project/sumaform/pull/1229
- https://github.com/uyuni-project/uyuni/pull/5240


ktsamis (approved):
The rest of the files already have the urls and we are now adding the rest as well as the definitions so they can use it. All seems good

PR #782, opened 2023-02-23
TEST ION: Add additional repo for server

meaksh:


juliogonzalez (commented):
Maybe you should add the one for the proxy to the proxy? I think at least one of the package will produce rebuilds for one of the packages used in the proxy.

PR #709, opened 2022-11-29
Adjust `hvm_disk_image` for recent sumaform changes

nodeg:
This PR adjusts the `hvm_disk_image` variables for the coming sumaform changes.

See
- https://github.com/uyuni-project/sumaform/pull/1197
- https://github.com/SUSE/spacewalk/issues/17238

PR #789, opened 2023-03-02
Adjust config for static nested VM hostname

nodeg:
After adding a static hostname for the nested VM in the infrastructure repository, I need to make adjustments here, too. Furthermore
- the SLES image is only needed on HEAD. Uyuni and the PR tests only use Leap 15.4
- the Leap 15.4 image is necessary everywhere since we use it for the normal KVM/virtualization tests (`minkvm_guests.feature`)
- we only use those on Uyuni/HEAD, no backports at the moment
- all minions that are properly set up with DNS/DHCP in our infrastructure have their proper hostname. All others have a generic `min-nested` hostname.

See https://github.com/uyuni-project/sumaform/pull/1263

### DIFF
--- a/terracumber_config/tf_files/SUSEManager-4.2-PRV.tf
+++ b/terracumber_config/tf_files/SUSEManager-4.2-PRV.tf
@@ -192,14 +192,9 @@ module "cucumber_testsuite" {
       additional_grains = {
         hvm_disk_image = {
           leap = {
-            hostname = "leap-salt-migration"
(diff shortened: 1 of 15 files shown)


ktsamis (approved):
Looks good, if we don't need the sles let's remove it.

nodeg:
We do not need it on for everything Uyuni related. I added it with https://github.com/SUSE/susemanager-ci/pull/709/files.

PR #793, opened 2023-03-07
Add salt bundle for centos7 on suma42

maximenoel8:
salt is not supported anymore for centos7. We need to use venv-salt.
Update suma4.2 dev acceptance terraform file to use salt-bundle during sumaform deployment.


srbarrios (approved):
Good catch thanks!

PR #807, opened 2023-03-30
BV: Make minions/clients selectable

maximenoel8:
**What does this PR do?**

Add a new parameter for build validation allowing to disable some nodes during a run.
Build with this new parameter : https://ci.suse.de/view/Manager/view/Manager-4.3/job/manager-4.3-qe-build-validation-parallel/build?delay=0sec

I attempted to use both the `ExtendedChoice` and `Active Choice` Jenkins plugins to build the list of nodes, but unfortunately, the latter requires a plugin that is not currently installed on CI.
 Active Choice allows to create more complex list ( with sub-list or sub-choices depending on a first choice). It also add a Filter feature. To describe the parameters, Active Choice is using Groovy script. The limit is the script will be executed on Jenkins Master and so we will not have access to our test environment.

filterActiveChoice

I was unable with both plugins to use a file to describe the node list. Both plugins require to have this file on Jenkins Master. Since it is not possible to find this file in Jenkins agents (workers), we would need to manually upload and maintain it on the Jenkins master, which is not ideal. To address this, I have directly included the list description in the Jenkins parameter file.

**How does it work ?**

To disable the minions, change need to be made in two places during the pipeline: Sync repositories stage and the client stage. To accomplish this, I will use an environment variable with the disabled nodes and call the unset command during the repository sync feature call. For example, if all `sles15sp3_ssh_minion`, `sles15sp3_minion`, and `sles15sp3_client` nodes are disabled, `sles15sp3` will not be synchronized.

For the testing client stage, I will remove the disabled nodes from the nodeList, ensuring that stages for these clients will not be created. Since the minion list is now needed in two stages, I have created a new function getMinionList that retrieves the minion list from the Terraform plan list and Jenkins parameter, and creates a new node list with the information required for the stages.

**To be discuss**

**1 ) More nodes declared on Jenkins side** 
I have encountered a scenario where more nodes are declared in Jenkins than are actually deployed. Currently, I am only logging a message about this discrepancy. However, we should consider whether this should cause the pipeline to fail instead. What action should we take in this situation?

**2 ) ssh minion naming**
We are facing a naming inconsistency between our testsuite and cucumber bashrc, which is causing unnecessary complexity in our pipeline. While the naming conventions are similar, there is a difference in the use of underscores. This is leading to complexity in our pipeline scripts. We need to decide on a standardized naming convention to be used across both our testsuite and cucumber bashrc.

**3 ) sle and sles naming**

We also have a difference between sumaform and testsuite about sle and sles naming.

Issue: Fixes https://github.com/SUSE/spacewalk/issues/20856

### DIFF
--- a/jenkins_pipelines/environments/common/pipeline-build-validation.groovy
+++ b/jenkins_pipelines/environments/common/pipeline-build-validation.groovy
@@ -396,7 +396,7 @@ def getMinionList() {
     String[] moduleList = modules.split("\n")
     moduleList.each { lane ->
         def instanceList = lane.tokenize(".")
-        if (instanceList[1].contains('minion') or instanceList[1].contains('client')) {
+        if (instanceList[1].contains('minion') || instanceList[1].contains('client')) {
             nodeList.add(instanceList[1].replaceAll("-", "_").replaceAll("sshminion", "ssh_minion").replaceAll("sles", "sle"))
             envVar.add(instanceList[1].replaceAll("-", "_").replaceAll("sles", "sle").toUpperCase())
         }
--- a/jenkins_pipelines/environments/common/pipeline-build-validation.groovy
+++ b/jenkins_pipelines/environments/common/pipeline-build-validation.groovy
@@ -401,9 +401,9 @@ def getMinionList() {
             envVar.add(instanceList[1].replaceAll("-", "_").replaceAll("sles", "sle").toUpperCase())
         }
     }
-    println ("Minion list from jenkins ${params.declareMinionList}" )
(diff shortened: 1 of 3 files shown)


maximenoel8:
I also renamed `error` to `result_error` variable because it was creating an issue in Result stage.

srbarrios (approved):
It looks good to me.
About the questions exposed in the descriptions, let's consider them separately. But there are all valid points that I agree we need to fix/improve.

Bischoff (approved):

nodeg (approved):
LGTM. I agree with Oscar. Also, do not forget to squash your commits before merging.

PR #821, opened 2023-04-04
[BV] Add a wait during MU stage for SSH minion

maximenoel8:
**What does this PR do ?**

Ssh minion and minion are running in parallel now. 
We can create the minion channel only once but this channel is needed for both minion and ssh minion.
The add MU repositories stage will only be executed during the minion pipeline.
The problem is the ssh minion will skip the Add MU repositories stage and will create an activation key before the minion MU channel is created.

This PR adds a wait for SSH minion during Add MU repositories stage to be sure minion MU repositories are correctly synchronized before creating the key.
I use a new variable setup in getMinionList that will be updated when minion and client finish the MU repositories synchronization.

### DIFF
--- a/jenkins_pipelines/environments/common/pipeline-build-validation.groovy
+++ b/jenkins_pipelines/environments/common/pipeline-build-validation.groovy
@@ -290,7 +290,7 @@ def clientTestingStages() {
                             error("Custom channels and MU repositories synchronization failed with status code: ${res_sync_mu_repos}")
                         }
                     } else {
-                        dependOn "add_MU_${minion.replaceAll('ssh_minion', 'minion')}"
+                        dependsOn "add_MU_${minion.replaceAll('ssh_minion', 'minion')}"
                     }
                 }
             }
--- a/jenkins_pipelines/environments/common/pipeline-build-validation.groovy
+++ b/jenkins_pipelines/environments/common/pipeline-build-validation.groovy
@@ -290,7 +290,7 @@ def clientTestingStages() {
                             error("Custom channels and MU repositories synchronization failed with status code: ${res_sync_mu_repos}")
                         }
                     } else {
-                        dependOn "add_MU_${minion.replaceAll('sshminion', 'minion')}"
+                        dependOn "add_MU_${minion.replaceAll('ssh_minion', 'minion')}"
                     }
                 }
             }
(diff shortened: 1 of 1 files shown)


jenkins_pipelines/environments/common/pipeline-build-validation.groovy
```diff
@@ -261,6 +261,7 @@ def clientTestingStages() {
 
     //Get minion list from terraform state list command
     def minionList = getMinionList()
+    def mu_sync_status = minionList.MUSyncStatus
```
srbarrios, line 264: I'm fine with this PR.
But I think it would be good to make some adjust with the names of some variables.

For instance:
```
minionList.nodeList.each { minion ->
```
- If we are iterating from "nodeList" we should be having as item "node".
- If we already have "minionList", we would guess by its name that we can iterate from it the list of minions, but instead we must use an internal attribute for that purpose. Should we name this variable "minionList" differently, I propose to call it "nodesHandler" and so "getNodesHandler" by the nature of its content.

jenkins_pipelines/environments/common/pipeline-build-validation.groovy
```diff
@@ -272,8 +273,16 @@ def clientTestingStages() {
...
-                    if (!minion.contains('ssh')) {
+                stage("Add_MUs_${minion}") {
+                    if (minion.contains('ssh_minion')) {
+                        // SSH minion need minion MU channel. This section wait until minion finish creating MU channel
+                        def minion_name_without_ssh = minion.replaceAll('ssh_minion', 'minion')
+                        println "Waiting for the MU channel creation by ${minion_name_without_ssh} for ${minion}."
+                        waitUntil {
+                            mu_sync_status[minion_name_without_ssh]
+                        }
+                        println "MU channel available for ${minion} "
```
srbarrios, line 284: It might be good to extract all this as a method.
For instance:
`waitUntilRepoIsReady(minion)`

maximenoel8: I'm not sure it will be possible to create a function. The `mu_sync_status` variable is define for the function. If this variable is updated during clientTestingStages the waitUntilRepoIsReady method will probably not see it.

jenkins_pipelines/environments/common/pipeline-build-validation.groovy
maximenoel8: I update this naming to the pipeline

PR #822, opened 2023-04-05
[BV] Add a random wait for boostrap and smoke test

maximenoel8:
**What does this PR do ?**

During BV pipeline, some nodes bootstrap and smoke test will fail because of UI issues.
This UI issues could come from the server side or controller side.
In my tests, the bootstrap and smoke test were more spread out ( and less users ) because I was running all the stages ( add MU / activation Key / boostrap reposync). So my nodes were not bootstrapping all at the same time.

To avoid having all the nodes ( 30 ~ ) to create a session and try to bootstrap at the same time add a random sleep ranging from 0 to 30 seconds.

This solution is not tested yet. We don't know if it will improve stability.

**Other solution**
We could also create a resource when bootstrapping saying only [2-10] nodes can bootstrap at the same time.

This solution is still open to discussion.

### DIFF
--- a/jenkins_pipelines/environments/common/pipeline-build-validation.groovy
+++ b/jenkins_pipelines/environments/common/pipeline-build-validation.groovy
@@ -365,6 +365,7 @@ def clientTestingStages() {
                     if (params.confirm_before_continue) {
                         input 'Press any key to start bootstraping the clients'
                     }
+                    randomWait()
                     echo 'Bootstrap clients'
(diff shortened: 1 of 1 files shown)


Bischoff (approved):

Bischoff:
Merging it to unblock me. We might revert it later.

Bischoff:
The PR indeed smoothes the CPU usage a lot
image
but all clients are still red

If you wish we can revert

PR #823, opened 2023-04-05
[BV] Compatibility between run by stage or run by node

maximenoel8:
**What does this PR do ?**

- Move `must_boot_monitoring` and `must_boot_proxy` parameter to the bootstrap part only.
- Add three new variables `enable_proxy_stages` , `enable_monitoring_stages` and `enable_client_stages`. (to be discuss)

After discussion with Eric, I found out we have a different approach to BV pipeline.
On my side I will run this pipeline per proxy, monitoring and clients. Where I will run all the steps for Proxy => Monitoring => client.
Eric will run it by steps : add MU => add non MU =>  add Activation key => etc

I designed the pipeline to have all this steps unique for each clients.
In the current version, I will not run `Prepare Proxy` stage or `Prepare Monitoring Server` if the option `must_boot_proxy` or `must_boot_monitoring` is not set to true.
So if you want to add MU for proxy, you need to have `add MU` to true and `boot proxy` to true. But the boot proxy will fail because `Activation Key` and `Bootstrap Repo Sync` were not executed. 

The `boot proxy` and `boot monitoring` should not be mandatory to Add MU for proxy and monitoring.

In this PR I added a new variables to keep my logic but this logic could be remove if nobody is using it. The idea is to be able to run the pipeline per node ( proxy, monitoring and clients).

We need to make a choice : 

 **Choice 1 :**
 We don't run the pipeline per node and so I remove the new `enable_****` variables

**Choice 2 :** 
We have the two possibilities :
 -  run per step ( Add MU = > Add Key => Create Bootstrap = > Bootstrap )
 - run per node ( Bootstrap Proxy with all stages => Bootstrap Monitoring with all stages => Run client smoke test will all prerequis )


Bischoff (commented):
three more checkboxes ? it's becoming complicated...

maximenoel8:
I agree, but the way we are using the pipeline is also complicated ;-)

PR #830, opened 2023-04-06
Remove unused alias in uyuni-bv config file

nodeg:
This alias is unnecessary and prevents the deployment from working. We need a default provider.

```bash
jenkins@jenkins-worker:~/jenkins-build/workspace/uyuni-master-qe-build-validation/results/sumaform> terraform plan
╷
│ Error: Missing required argument
│ 
│ The argument "uri" is required, but was not set.
```


Bischoff (approved):
yes i should have reacted when you added it

PR #833, opened 2023-04-11
QE: switch uyuni-bv to uyuni-released

nodeg:
As discussed with @juliogonzalez and @srbarrios, we will use `uyuni-released`. When a new Uyuni version will be released and Julio prepared everything, we use a JSON to inject the new code changes from `uyuni-master` as we do for example in our SUMA BV with the MU/MIs.


juliogonzalez (approved):
LGTM, assuming you didn't miss any line with `uyuni-master` (this file is huge! :-D)

nodeg:
I did a search-and-replace :wink:

PR #852, opened 2023-05-03
Make sonarcloud PR variables detection more robust

cbosdo:
Don't fail if there is the string `ref` in the PR label.


mcalmer (approved):
nice

PR #853, opened 2023-05-03
Set up new openqa installation jobs

ktsamis:
This creates 2 pipelines, one for online and one for offline installation in openqa.

Essentially we are setting up a json file that contains the variables we want to override/send to openQA and then we use that json with running the script from the openqa-executor. 

The only actual variable that you can customize is the ISO_URL for now the default it the latest from https://download.suse.de/ibs/SUSE:/SLE-15-SP4:/Update:/QR:/TEST/images/iso/

Both these jobs are running from my fork currently here:

Online: https://ci.suse.de/view/Manager/view/Manager-qe/job/openqa-installation_TEST2/ (already have a ticket with infra to rename)
Offline: https://ci.suse.de/view/Manager/view/Manager-qe/job/manager-4.3-qe-openqa-offline-installation/

### DIFF
--- a/jenkins_pipelines/environments/common/openqa.groovy
+++ b/jenkins_pipelines/environments/common/openqa.groovy
@@ -3,19 +3,19 @@ def run(params) {
         stage('Installation') {
             script {
               if (params.Installation_Type == 'online') {
-                sh "ssh -v -o StrictHostKeyChecking=no root@openqa-executor.mgr.suse.de; cd /root/openqa-suma-installation; cp params-run-installation-43_online.json params-run-installation-43.json;"
+                sh "ssh -v -o StrictHostKeyChecking=no root@openqa-executor.mgr.suse.de && cd /root/openqa-suma-installation && cp params-run-installation-43_online.json params-run-installation-43.json;"
               } else {
-                sh "ssh -v -o StrictHostKeyChecking=no root@openqa-executor.mgr.suse.de; cd /root/openqa-suma-installation; cp params-run-installation-43_full.json params-run-installation-43.json;"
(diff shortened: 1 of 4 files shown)


aaannz (commented):
Looks good to me.

To future proof it little more you can also add SUMA version as a Jenkins parameter and depending on it modify proper json file.

E.g. having parameter `SUMA_version` will in parametrize stage try to edit `params-run-installation-${params.SUMA_version}-${params.Installation_Type}.json` so when we will want this for future SUMA we do not have to edit pipeline, just create new json file on executor and new json job with correct version.

With `Installation_Type` parameterized, you can also IMO get rid of the branching in parameterize stage and just use one command with online/offline replaced by installation type.

ktsamis:
> Looks good to me.
> 
> To future proof it little more you can also add SUMA version as a Jenkins parameter and depending on it modify proper json file.
> 
> E.g. having parameter `SUMA_version` will in parametrize stage try to edit `params-run-installation-${params.SUMA_version}-${params.Installation_Type}.json` so when we will want this for future SUMA we do not have to edit pipeline, just create new json file on executor and new json job with correct version.
> 
That is possible but I want to reduce it as an option because if you add it to jenkins people will be able to edit it in each job run, if someone accidentally clicks on it then they will trigger something completely different from the 4.3 named job  but they will be running 4.2 for example. I would rather we copy over and edit the current job in each new jenkins pipeline with the hardcoded version to be on the safe side.


> With `Installation_Type` parameterized, you can also IMO get rid of the branching in parameterize stage and just use one command with online/offline replaced by installation type.

agreed, this was kind of the reverse of the above point, to have a common pipeline for each version between offline and online. But I will fix that now to simplify and have one approach, with less parameters.

PR #854, opened 2023-05-04
Refactor BV to have stage always apparent

maximenoel8:
### What does this PR change ?
Move the if condition inside the stage definition to have the stage always visible even when skip.
Split `Prepare stage` into stages corresponding to Jenkins parameters for Proxy and Monitoring.
Add node name into generic stage name.

### DIFF
--- a/jenkins_pipelines/environments/common/pipeline-build-validation.groovy
+++ b/jenkins_pipelines/environments/common/pipeline-build-validation.groovy
@@ -74,53 +74,60 @@ def run(params) {
                 }
             }
 
-            if (params.enable_proxy_stages) {
-                stage('Prepare Proxy') {
-                    if (params.must_add_MU_repositories) {
-                        echo 'Add proxy MUs'
-                        if (params.confirm_before_continue) {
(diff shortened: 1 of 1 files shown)


srbarrios (approved):
It's a matter of taste. I'm fine approving this.

Bischoff:
> It's a matter of taste. I'm fine approving this.

If some stages are apparent in a run, and others not, then you view only latest run - jenkins cannot cope with changing columns.

For me it's important to see all runs and be able to compare them. So yes it's a matter of tastes, but it's also conditioned by the way of working.

Bischoff (approved):

maximenoel8:
Maybe at the end we just need one drop list parameter with Eric / Maxime / Oscar option ;-)

maximenoel8:
After my testing, if some stages are failing, it will not show all the stages ( like all the other pipeline ). And so it will not display the previous run correctly in the main Jenkins view. But like I explain to Eric, the best way to read the parameter is in Blue Ocean.

PR #876, opened 2023-05-30
[BV] Correctly fail add MU ssh minion stage if minion stage fails

maximenoel8:
## What does this PR ?

SSH Minion and minion both need the minion MU repository when testing in BV. But we only can sync MU repository once. To solve this situation, only minion stage is creating the MU channel and ssh minion will use it.
Because ssh minion and minion are running in parallel, during add MU stage ssh minion are using a Jenkins waitUntil to wait the minion add MU stage to finish.

If the add MU minion stage fails, the add MU minion stage will wait forever that the MU minion stage succeed. 

To solve this situation, I changed `mu_sync_status[node]`  variable to string so it can store three states : 
 - `NOSYNC`, set when creating the variable
 - `SYNC`, set after minion add MU stage succeed
 - `FAIL`, set if minion add MU stage failed.

The waitUntil will then wait for `SYNC` status or `FAIL`. And a condition after that will fail the stage if the status was `FAIL`

### DIFF
--- a/jenkins_pipelines/environments/common/pipeline-build-validation.groovy
+++ b/jenkins_pipelines/environments/common/pipeline-build-validation.groovy
@@ -294,9 +294,9 @@ def clientTestingStages(capybara_timeout, default_timeout) {
                         def minion_name_without_ssh = node.replaceAll('ssh_minion', 'minion')
                         println "Waiting for the MU channel creation by ${minion_name_without_ssh} for ${node}."
                         waitUntil {
-                            if (mu_sync_status[minion_name_without_ssh] = "SYNC") {
+                            if (mu_sync_status[minion_name_without_ssh] == 'SYNC') {
                                 return true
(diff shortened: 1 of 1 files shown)


jenkins_pipelines/environments/common/pipeline-build-validation.groovy
```diff
@@ -294,28 +294,34 @@ def clientTestingStages(capybara_timeout, default_timeout) {
                         def minion_name_without_ssh = node.replaceAll('ssh_minion', 'minion')
                         println "Waiting for the MU channel creation by ${minion_name_without_ssh} for ${node}."
                         waitUntil {
-                            mu_sync_status[minion_name_without_ssh]
+                            (mu_sync_status[minion_name_without_ssh] == 'SYNC' || mu_sync_status[minion_name_without_ssh] == 'FAIL')
```
srbarrios, line 297: I guess you can use only one comparision until not UNSYNC. To simplify?

srbarrios (approved):

PR #884, opened 2023-06-08
Disable mirror in Uyuni-Master-K3S

srbarrios:
The PRV mirror is giving problems, let's temporary disable it in our K3S Uyuni Test suite


srbarrios:
I think we can close this PR.

PR #892, opened 2023-06-14
Use hyperion for running acceptance tests at PR for SUMA 4.3

jordimassaguerpla:
This removes environments 5 and 6 from the uyuni-prs-ci-test . Environments 5 and 6 use hyperion. This way, we can use hyperion for suma 4.3 tests.

related to https://github.com/SUSE/spacewalk/issues/21471


maximenoel8 (approved):

maximenoel8:
If I'm correct you remove also env 7 and 8 with this change right ?

PR #893, opened 2023-06-14
Define terraform conf files for SUMA 4.3 Acceptance tests PR testing

jordimassaguerpla:
Based on what we have for Uyuni. Let's use hyperion.

This goes after https://github.com/SUSE/susemanager-ci/pull/892

Related to https://github.com/SUSE/spacewalk/issues/21471

PR #895, opened 2023-06-15
Reserver hyperion for 4.3. We will install a new hyp later.

jordimassaguerpla:


jordimassaguerpla:
@maximenoel8 : This is just so that we do not use env5 and env6 for testing uyuni. We will revert this later.

jordimassaguerpla:
This way, there won't be any danger on using https://github.com/SUSE/susemanager-ci/pull/893/files

jordimassaguerpla:
Superseded by https://github.com/SUSE/susemanager-ci/pull/893/

PR #902, opened 2023-06-27
Fix Code Coverage pipeline by using previous TF file

srbarrios:


srbarrios:
Merging directly, Code coverage pipeline still a PoC I'm working on.

PR #907, opened 2023-07-06
Fix duplicate MAC addresses in the Uyuni BV

nodeg:
We had duplicated MAC addresses for our openSUSE 15.x ARM minions and SSH minions in NUE and PRV.


Bischoff (approved):
my bad, from commits c0e37ed7 and b9b76fa0

PR #910, opened 2023-07-24
Remove overdrive and arm clients from BV

maximenoel8:
overdrive machines are not available anymore. We need to remove it from BV because it will fail all the BV deployment.


srbarrios:
Are no available anymore or are just down for a while until they are moved to the new place?
If they will be back, I would comment instead of remove.

PR #913, opened 2023-07-28
[BV]Harmonize capybara timeout / monitoring persistent

maximenoel8:
Harmonize Capybara timeout to be setup for all the rake calls.
Remove random wait from Add MU / Add Non MU and create Activation Key ( I was able to reproduce the same behavior even with the wait ). It was more likely trigger by a too short timeout during login.
Add a persistent file state to manage the MUs shared by monitoring and sles15sp4_minion.
Indeed they are using the same MUs, you can't add twice the same MU.
I created a file so we keep the monitoring state between run.

### DIFF
--- a/jenkins_pipelines/environments/common/pipeline-build-validation.groovy
+++ b/jenkins_pipelines/environments/common/pipeline-build-validation.groovy
@@ -46,6 +46,10 @@ def run(params) {
 
             stage('Deploy') {
                 if (params.must_deploy) {
+                    // Delete persistent monitoring state
+                    // Explanation : monitoring is using the same MU than one of the sles minion, we can't add twice the MUs
+                    // We are using a persistent file to show monitoring MU has been added
+                    sh "rm -f ${env.resultdir}/sumaform/monitoring"
                     // Generate json file in the workspace
                     writeFile file: 'custom_repositories.json', text: params.custom_repositories, encoding: "UTF-8"
                     // Run Terracumber to deploy the environment
@@ -144,6 +148,10 @@ def run(params) {
                             res_sync_mu_repos = sh(script: "./terracumber-cli ${common_params} --logfile ${resultdirbuild}/testsuite.log --runstep cucumber --cucumber-cmd '${env.exports} cd /root/spacewalk/testsuite; rake cucumber:build_validation_wait_for_custom_reposync'")
                             echo "Custom channels and MU repositories synchronization status code: ${res_sync_mu_repos}"
                             sh "exit \$(( ${res_mu_repos}|${res_sync_mu_repos} ))"
+                            // Add persistent monitoring state
+                            // Explanation : monitoring is using the same MU than one of the sles minion, we can't add twice the MUs
+                            // We are using a persistent file to show monitoring MU has been added
+                            sh "touch ${env.resultdir}/sumaform/monitoring"
                         }
                     }
                     stage('Add Activation Keys Monitoring') {
@@ -312,6 +320,9 @@ def clientTestingStages() {
     //Get minion list from terraform state list command
     def nodesHandler = getNodesHandler()
     def mu_sync_status = nodesHandler.MUSyncStatus
+    // Check persistent monitoring state
+    // Explanation : monitoring is using the same MU than one of the sles minion, we can't add twice the MUs
+    // We are using a persistent file to show monitoring MU has been added
     def monitoring_bootstrap_output = sh(script: "test -e ${env.resultdir}/sumaform/monitoring && echo 'true' || echo 'false'", returnStdout: true).trim()
     // Construct a stage list for each node.
     nodesHandler.nodeList.each { node ->
--- a/jenkins_pipelines/environments/common/pipeline-build-validation.groovy
+++ b/jenkins_pipelines/environments/common/pipeline-build-validation.groovy
@@ -59,7 +59,7 @@ def run(params) {
             }
 
             stage('Sanity check') {
-                sh "./terracumber-cli ${common_params} --logfile ${resultdirbuild}/testsuite.log --runstep cucumber --cucumber-cmd 'cd /root/spacewalk/testsuite; ${env.exports} rake cucumber:build_validation_sanity_check'"
+//                sh "./terracumber-cli ${common_params} --logfile ${resultdirbuild}/testsuite.log --runstep cucumber --cucumber-cmd 'cd /root/spacewalk/testsuite; ${env.exports} rake cucumber:build_validation_sanity_check'"
             }
 
             stage('Run core features') {
--- a/jenkins_pipelines/environments/common/pipeline-build-validation.groovy
+++ b/jenkins_pipelines/environments/common/pipeline-build-validation.groovy
@@ -312,7 +312,7 @@ def clientTestingStages() {
     //Get minion list from terraform state list command
     def nodesHandler = getNodesHandler()
     def mu_sync_status = nodesHandler.MUSyncStatus
-
+    def monitoring_bootstrap_output = sh(script: "test -e ${env.resultdir}/sumaform/monitoring && echo 'true' || echo 'false'", returnStdout: true).trim()
     // Construct a stage list for each node.
     nodesHandler.nodeList.each { node ->
         tests["${node}"] = {
@@ -336,7 +336,7 @@ def clientTestingStages() {
                         } else {
                             println "MU channel available for ${node} "
                         }
-                    } else if (node == "${params.monitoring_sle_version}_minion" && new File("${env.resultdir}/sumaform/monitoring").exists()) {
+                    } else if (node == "${params.monitoring_sle_version}_minion" && monitoring_bootstrap_output == 'true') {
                         mu_sync_status[node] = 'SYNC'
                     } else {
                         if (params.confirm_before_continue) {
--- a/jenkins_pipelines/environments/common/pipeline-build-validation.groovy
+++ b/jenkins_pipelines/environments/common/pipeline-build-validation.groovy
@@ -63,7 +63,7 @@ def run(params) {
             }
 
             stage('Sanity check') {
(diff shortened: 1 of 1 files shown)


jenkins_pipelines/environments/common/pipeline-build-validation.groovy
```diff
@@ -1,11 +1,15 @@
 def run(params) {
     timestamps {
+        //Capybara configuration
+        def capybara_timeout = 60
```
srbarrios, line 4: No please. You are overwritting and not taking into account the jenkins values passed by parameter when running the BV.
We should not have it here as constants. Starting by the fact that not all environments performs the same way.

jenkins_pipelines/environments/common/pipeline-build-validation.groovy
```diff
@@ -144,6 +148,10 @@ def run(params) {
                             res_sync_mu_repos = sh(script: "./terracumber-cli ${common_params} --logfile ${resultdirbuild}/testsuite.log --runstep cucumber --cucumber-cmd '${env.exports} cd /root/spacewalk/testsuite; rake cucumber:build_validation_wait_for_custom_reposync'")
                             echo "Custom channels and MU repositories synchronization status code: ${res_sync_mu_repos}"
                             sh "exit \$(( ${res_mu_repos}|${res_sync_mu_repos} ))"
+                            // Add persistent monitoring state
+                            // Explanation : monitoring is using the same MU than one of the sles minion, we can't add twice the MUs
+                            // We are using a persistent file to show monitoring MU has been added
+                            sh "touch ${env.resultdir}/sumaform/monitoring"
```
srbarrios, line 154: If we want to use lock files, better if we do inside a folder with a self explanatory name, and a file name that ends in .lock to make it clear.

srbarrios (commented):
In our BV pipeline, at least before the big refactor that was hard to review, we had Jenkins parameters to provide when running the pipeline to setup a value for the DEFAULT_TIMEOUT, now this seems not possible, please let's have that back.

srbarrios (commented):
I can also notice that previously we could scale in clients without the need to make change in the pipeline, now, it seems we have some stages like for retail that specifically declare variable with OS names. This will cause a harder maintenance when we add more supported clients. Let' please re-think that.

maximenoel8:
I never change the retail part. Let's not mix everything at the same time. This PR is to improve BV stability.

jenkins_pipelines/environments/common/pipeline-build-validation.groovy
```diff
@@ -46,6 +46,10 @@ def run(params) {
 
             stage('Deploy') {
                 if (params.must_deploy) {
+                    // Delete persistent monitoring state
+                    // Explanation : monitoring is using the same MU than one of the sles minion, we can't add twice the MUs
+                    // We are using a persistent file to show monitoring MU has been added
+                    sh "rm -f ${env.resultdir}/sumaform/monitoring"
```
srbarrios, line 52: I would store/remove it inside the testsuite folder.

jenkins_pipelines/environments/common/pipeline-build-validation.groovy
```diff
@@ -312,7 +320,10 @@ def clientTestingStages(capybara_timeout, default_timeout) {
     //Get minion list from terraform state list command
     def nodesHandler = getNodesHandler()
     def mu_sync_status = nodesHandler.MUSyncStatus
-
+    // Check persistent monitoring state
+    // Explanation : monitoring is using the same MU than one of the sles minion, we can't add twice the MUs
+    // We are using a persistent file to show monitoring MU has been added
+    def monitoring_bootstrap_output = sh(script: "test -e ${env.resultdir}/sumaform/monitoring && echo 'true' || echo 'false'", returnStdout: true).trim()
```
srbarrios, line 326: Why don't handle this from inside the test framework?
We can only need to check if the MU has been added, if so, we skip the step.

That would very much simplify the complexity of the pipeline, moving this responsibility to the test framework.

maximenoel8: I'm not sure to understand how to move this to the test framework. 
Do you mean we you should always run the stage on the pipeline side and add steps int the testsuite that check if MU exist before adding it ? It would change drastically our testsuite on this side. For now we are failing if you try to add an existing MU. 
This behavior/ issue is just the consequence of reusing the same MU for monitoring and sles15sp4.
For me it's more a pipeline behavior and only the Jenkins worker can store if the monitoring call has been made before and no new deployment were executed again. And only the Jenkins worker can skip Rake call.

But I'm open to other idea if you have a better way to manage this. But you need to understand why we need it in the first place.

vandabarata:
```ruby
  desc 'Generate feature to add Maintenance Update Channels and repositories for all the clients'
  task :generate_add_maintenance_update_repository_features do
    clients = Rake::FileList['features/build_validation/init_clients/*.feature']
    clients.reject! { |client| client.include?('ssh') }
    clients.reject! { |client| client.include?('monitoring_server') }
```
we should already be doing that, since rake should skip over the monitoring server. Something needs to be fixed here if that's not the case :thinking: 
This is in our Rakefile, for reference.

srbarrios: @maximenoel8 Maybe I missed something, but we already have two ways to prevent duplication, one as mentioned by Vanda is this reject method we have in some rake tasks.
The other is this code we have in the test framework:
```
When(/^I create the MU repositories for "([^"]*)"$/) do |client|
  repo_list = $custom_repositories[client]
  next if repo_list.nil?

  repo_list.each do |_repo_name, repo_url|
    unique_repo_name = generate_repository_name(repo_url)
    if repository_exist? unique_repo_name
      log "The MU repository #{unique_repo_name} was already created, we will reuse it."
    else
      content_type = deb_host?(client) ? 'deb' : 'yum'
      steps %(
        When I follow the left menu "Software > Manage > Repositories"
        And I follow "Create Repository"
        And I enter "#{unique_repo_name}" as "label"
        And I enter "#{repo_url.strip}" as "url"
        And I select "#{content_type}" from "contenttype"
        And I click on "Create Repository"
        Then I should see a "Repository created successfully" text or "The repository label '#{unique_repo_name}' is already in use" text
        And I should see "metadataSigned" as checked
      )
    end
  end
end
```

Note this part:
```
    if repository_exist? unique_repo_name
      log "The MU repository #{unique_repo_name} was already created, we will reuse it."
```

maximenoel8:
```
if repository_exist? unique_repo_name
      log "The MU repository #{unique_repo_name} was already created, we will reuse it."
```
could solve the issue. I need to test if it works correctly.

To explain the issue the easiest way, let's take sles15 sp4. For sles15sp4 we have 4 different types of minion ( I consider monitoring a minion in this example to simplify. ): 
 - sles15sp4_client
 - sles15sp4_ssh_minion
 - sles15sp4_minion
 - monitoring_server 

Then we have two type of MU repositories : 
 - sles15sp4_client_MU
 - sles15sp4_minion_MU

You can see we have some math issues, so now how to link this MUs : 
 - sles15sp4_client_MU :
   -  sles15sp4_client
 - sles15sp4_minion_MU : 
   - sles15sp4_ssh_minion
   - sles15sp4_minion
   - monitoring_server 

This was not an issue before because we were adding all the MU at the same time so we didn't need any order. But now this stage is split by minion and create the question during which node call do you add the sles15sp4_minion_MU ?

For all the other sles versions, slesX_minion_MU will be added during the client stages when calling the slesX_minion add MU stage and the slesX_ssh_minion will wait for slesX_minion to be done.

But for sles15sp4_minion_MU, we need it for monitoring bootstrap. And monitoring is call in a separate stage before the client stages. During this stage I will call add MU sles15sp4_minion because we don't have specific MU for monitoring.
Because the MU was added during the sequential monitoring stage, it should not be added again during this client stages. That's why I have this skip for sles15sp4_minion add MU stage.

And now, why do I need a persistent state ? Currently, I'm checking if monitoring is enable to skip sles15sp4 add MU stage. But you can call monitoring stages and client stages in two different Runs. By doing that, adding sles15sp4 MU during minion is failing.

Now you have the full picture.

The Rake doesn't have impact in this scenario but the Mu skip could.

jenkins_pipelines/environments/common/pipeline-build-validation.groovy
maximenoel8: It was asked by the team to remove some of the parameters we have in the configuration. capybara timeout is not really something we change. If we want a specific timeout depending on platform, it should be added into manager-4.3-qe-build-validation-PRV

maximenoel8:
The skip MU works correctly but the Add MU stage also create a channel. And we don't check for this one : 
```
Scenario: Add the custom child channel for sle15sp4_minion                        # features/build_validation/add_MU_repositories/sle15sp4_minion_add_maintenance_update_repositories.feature:10
This scenario ran at: 2023-07-31 03:52:09 +0200
When I follow the left menu "Software > Manage > Channels"                      # features/step_definitions/navigation_steps.rb:354
WARN: Step ends with an ajax transition not finished, let's wait a bit!
And I follow "Create Channel"                                                   # features/step_definitions/navigation_steps.rb:301
And I enter "Custom Channel for sle15sp4_minion" as "Channel Name"              # features/step_definitions/navigation_steps.rb:237
And I enter "custom_channel_sle15sp4_minion" as "Channel Label"                 # features/step_definitions/navigation_steps.rb:237
And I select the parent channel for the "sle15sp4_minion" from "Parent Channel" # features/step_definitions/navigation_steps.rb:203
And I enter "Custom channel" as "Channel Summary"                               # features/step_definitions/navigation_steps.rb:237
And I click on "Create Channel"                                                 # features/step_definitions/navigation_steps.rb:273
Then I should see a "Channel Custom Channel for sle15sp4_minion created" text   # features/step_definitions/navigation_steps.rb:587
      Text 'Channel Custom Channel for sle15sp4_minion created' not found (RuntimeError)
      ./features/step_definitions/navigation_steps.rb:588:in `/^I should see a "([^"]*)" text$/'
      features/build_validation/add_MU_repositories/sle15sp4_minion_add_maintenance_update_repositories.feature:18:in `I should see a "Channel Custom Channel for sle15sp4_minion created" text'
```

The solution could be to create a custom channel for monitoring.

PR #914, opened 2023-07-30
Disable auto create bootstrap repository during BV

maximenoel8:
Disable auto create bootstrap repository during BV

Required : https://github.com/uyuni-project/sumaform/pull/1357


mcalmer (approved):

terracumber_config/tf_files/SUSEManager-4.2-build-validation-NUE.tf
```diff
@@ -168,6 +168,7 @@ module "server" {
   publish_private_ssl_key        = false
   use_os_released_updates        = true
   disable_download_tokens        = false
+  disable_auto_bootstrap         = true
```
srbarrios, line 171: I would say the feature of auto bootstrap was only delivered as part of 4.3
But.. I also don't think this appended property will hurt in the rhn.conf file

PR #916, opened 2023-08-01
Move sync between minion and ssh minion during the create bootstrap

maximenoel8:
## SSH minion activation key
To create the SSH minion Activation Key, we need to have the minion Custom Channel available
I created `CustomChannelStatus` variable to check if the Add MU (and|or) Add Non MU stage are finished for the related minion.

## SSH minion bootstrap
SSH minion needs the bootstrap repository during the bootstrap. 
This bootstrap repository is created by the minion version of themself during create bootstrap repository stage.
Now that auto create bootstrap repository is disable during BV, it has to be trigger by minion.
It's a good think because now we are sure that the bootstrap repository have correctly all the custom channel ready when created.
Before that, some ssh minions were probably bootstrap without having the custom channel...

Create a new variable `BootstrapRepositoryStatus` to manage bootstrap repo creation status.
We don't need to sync during add MU stage anymore because we are syncing at create bootstrap stage.
This change will also fix this issue : https://github.com/SUSE/spacewalk/issues/22146

## Timeout increase

In our testsuite we have 500 secs timeout. This testsuite Timeout was breaking before the 500 sec Timeout.

## Issues
Related to : 
- https://github.com/SUSE/spacewalk/issues/22146

### DIFF
--- a/jenkins_pipelines/environments/common/pipeline-build-validation.groovy
+++ b/jenkins_pipelines/environments/common/pipeline-build-validation.groovy
@@ -312,6 +312,7 @@ def clientTestingStages() {
     //Get minion list from terraform state list command
     def nodesHandler = getNodesHandler()
     def mu_sync_status = nodesHandler.MUSyncStatus
+    def bootstrap_repository_status = nodesHandler.BootstrapRepositoryStatus
     // Construct a stage list for each node.
     nodesHandler.nodeList.each { node ->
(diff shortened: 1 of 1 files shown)


elariekerboull (approved):

srbarrios (approved):
Reading in diagonal.. but lgtm, let's give a try.

PR #918, opened 2023-08-04
Hexagon pipeline: Switch from SLES to openSUSE Leap

mbussolotto:
Hexagon pipeline: Switch from SLES to openSUSE Leap


nodeg (approved):
LGTM.

PR #924, opened 2023-08-14
Correctly add liberty9 back to controller

maximenoel8:
Add liberty9 to controller


nodeg (changes requested):
Small remark: I would also enable the SSH minion again.

nodeg (approved):
LGTM. Thank you, Maxime.

PR #927, opened 2023-08-16
overdrive4 is reconnected again

Bischoff:


nodeg:
What about the other Uyuni BV file?

Bischoff:
> What about the other Uyuni BV file?

The BV files in NUE rely on overdrive3, which is not up again yet.

PR #943, opened 2023-08-30
add ubuntu image to hexagon

mbussolotto:
add ubuntu image to hexagon


cbosdo:
How comes you add the image but not the minion?

PR #945, opened 2023-09-04
use opensuse in hexagon tests

mbussolotto:
use opensuse in hexagon tests


mbussolotto:
@cbosdo , k3s tests are already using opensuse. This PR should fix at least the `Update activation keys` test failure (they happens just in podman and not in k3s)

cbosdo (approved):
Remember that `master` still has issues with build host using leap

PR #947, opened 2023-09-08
Revert pxeboot-minion from Leap to SLES15 SP4

nodeg:
Due to missing support for openSUSE Leap as underlying base OS for the pxeboot-minion, we revert back to SLES15 SP4.


vandabarata (approved):

ktsamis (approved):
Sad to do this but ok :(

PR #950, opened 2023-09-12
Disable java_debugging in BV environments

srbarrios:
This is just a proposal to speed up our BV tests.
Currently we have extra java logs printed out by default. With this PR, we will disable those extra debugging logs.


nodeg (approved):
LGTM.

PR #964, opened 2023-09-18
Use openstack minimal-vm images as they have no firstboot wizard

cbosdo:


jordimassaguerpla (approved):

cbosdo:
The images don't exist on 15.5

PR #972, opened 2023-09-29
New BV pipeline parameter to provide a list of MI IDs and generate the JSON on the fly

srbarrios:
Addressing https://github.com/SUSE/spacewalk/issues/22506

This PR introduce a new parameter in our BV pipeline (only regular pipeline, not AWS).
- The name of this parameter is "mi_ids" and it expect a list of MI identifiers separated by comma without whitespaces.
- If this parameter is not empty, it will overwrite the JSON provided on the parameter "custom_repositories".
- We moved the script to generate the JSON from the spacewalk/uyuni repository to susemanager-ci repository, that way we can call it without depending on a controller deployment.
- In order to call OSC command, we are using an extra Jenkins node in the pipeline, for all the rest we keep using the same node.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,2 @@
+.idea
+susemanager-ci.iml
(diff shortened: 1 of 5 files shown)


ktsamis (approved):

nodeg:
Please update the Python script once https://github.com/uyuni-project/uyuni/pull/7653 will get merged.

PR #974, opened 2023-10-05
Update to leap 15.5: kvm isos in K3s

jordimassaguerpla:
In master is working, so let's apply it here as well.


jordimassaguerpla:
fixes https://github.com/SUSE/spacewalk/issues/22689

cbosdo (approved):

PR #979, opened 2023-10-05
Update to Leap 15.5 SUSEManager-Head-NUE

jordimassaguerpla:
fixes https://github.com/SUSE/spacewalk/issues/22705


nodeg:
The current change you did is only for the nested VM (Salt bundle migration tests) that will be removed soon. The Salt bundle migration tests will move from CI to BV.

PR #981, opened 2023-10-06
Fix Undelivered Mail Returned to Sender error

nodeg:
Recently, we saw the following issue on our `galaxy-ci` mailing list:

```bash
Diagnostic-Code: smtp; 554 5.7.1 <galaxy-ci@suse.de>: Sender address rejected:
    We don't accept our own addresses as sender address on this host!
```


juliogonzalez (approved):
LGTM. Assuming you used `sed` for this, so you are not missing any `.tf` file :-)

PR #898, opened 2023-06-16
Create a PR main.tf template and use tfvars to specify test envs

maximenoel8:
## TESTING-DONT-REVIEW-pipeline-pull-request-light.groovy is for testing purpose. DON'T REVIEW
**DON"T REVIEW TESTING-DONT-REVIEW-pipeline-pull-request-light.groovy**

This PR is a revamp of all main.tf for the PR testing pipelines to manage the differences in tfvars files and have one unique main.tf template common for all the PR testing environments.

Related to https://github.com/SUSE/spacewalk/issues/21471

PR #990, opened 2023-10-12
Use normalize name for rh and deb like minion in uyuni PR testing

maximenoel8:
I'm currently working on a main.tf template to be use accross all PR testing ( NUE / PRV and uyuni / manager ).
We are already using normalize name for suse minions but not for rh and deb like minions ( those normalize names are already used in CI ) .

This PR will normalize rh and deb minion name.

Require : https://gitlab.suse.de/galaxy/infrastructure/-/merge_requests/813
Related to : https://github.com/SUSE/susemanager-ci/pull/898

### DIFF
--- a/terracumber_config/tf_files/Uyuni-PR-tests-env1.tf
+++ b/terracumber_config/tf_files/Uyuni-PR-tests-env1.tf
@@ -256,7 +256,7 @@ module "cucumber_testsuite" {
     }
     redhat-minion = {
       image = "rocky8o"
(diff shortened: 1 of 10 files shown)


maximenoel8:
The check is failing but I don't really understand the issue.

nodeg (approved):
LGTM. Thank you, Maxime.

nodeg:
Regarding the failed tests, we could ping Can, but I also see no reason why the test failed with this simple renaming change.

PR #993, opened 2023-10-12
Fix validator for all .tf files

cbbayburt:
Fixes:
 - Add missing variable declarations (all mandatory variables must be set in env)
 - Remove additional `libvirt` provider definitions
 - Consider "commented out" lines in regexps


nodeg (approved):
LGTM. Thank you, Can.

PR #998, opened 2023-10-16
Try to work around dependency issue

Bischoff:
This might work around issue https://github.com/uyuni-project/uyuni/issues/7637 that prevents from deploying Debian 10 and Ubuntu 20.04.

If that does not work out, we'll simply disable them for the time being.


nodeg (approved):
LGTM.

PR #1003, opened 2023-10-23
Remove monitoring stage from QE SLE testsuite since we do not deploy monitoring server

ktsamis:
The possibility to check the stage is still there, even if it's not used by default. Checking the stage and running it would cause an error:

```
 Don't know how to build task 'cucumber:build_validation_add_maintenance_update_repositories_monitoring_server' (See the list of available tasks with `rake --tasks`)
```

Removing it should prevent this from happening.


nodeg (approved):

ktsamis:
I can confirm that removing the stage causes no other error and running the pipeline with parameters this doesn't appear at all as an option. So we should be good to merge.

PR #1017, opened 2023-11-15
Use 15.5 in Hexagon Pipeline

jordimassaguerpla:
fixes https://github.com/SUSE/spacewalk/issues/22702

We need to keep 15.4 image because the controller is 15.4


mbussolotto (approved):
LGTM

PR #1041, opened 2023-12-01
Add support for s390 VMs

Bischoff:
This PR adds support for s390 VMs, but leaves them disabled for how long we do not have a working production environement.


nodeg:
The CI does not find your provider.

ktsamis:
usually for sumaform, we use a symlink to link to the provider I think, that might be it

Bischoff:
> usually for sumaform, we use a symlink to link to the provider I think, that might be it

No, that symlink is done by terracumber, which is apparently not involved in the CI tests for susemanager-ci.

PR #1045, opened 2023-12-14
The bundle is now available on Ubuntu too

Bischoff:
Remove an old FIXME

This will hopefully remove that error:
```
[2023-12-14T14:32:22.663Z] ^[[0m^[[1mmodule.cucumber_testsuite.module.debian-minion.module.minion.module.host.null_resource.provisioning[0] (remote-exec):^[[0m ^[[0mpkg_resources.DistributionNotFound: The 'contextvars' distribution was not found and is required by salt
```


nodeg:
Are these the only `.tf` files?

Bischoff:
> Are these the only `.tf` files?

according to my grep, yes.

nodeg (approved):

PR #1050, opened 2024-01-03
Set external disks for repository and database

srbarrios:
Depends on https://github.com/uyuni-project/sumaform/pull/1454

This PR aims to setup a bigger disk size for pgsql partition, as otherwise we had problems on CI tests, when we run the backup/restore test.

**Will those disk sizes make sense to you?**

I set up the official minimal size, but maybe we can work with less in CI?
image

```
suma-43-srv:/var/spacewalk # du -h -d0
54G	.
suma-43-srv:/var/lib/pgsql # du -h -d0
6,5G	.
```

Note: In Uyuni, I expect more, as we synchronize Leap (even if we kill it later) + SLES15SP4. But I don't have a working CI testsuite to check it.

### DIFF
--- a/terracumber_config/tf_files/SUSEManager-4.3-NUE.tf
+++ b/terracumber_config/tf_files/SUSEManager-4.3-NUE.tf
@@ -135,6 +135,8 @@ module "cucumber_testsuite" {
         vcpu = 8
         memory = 32768
       }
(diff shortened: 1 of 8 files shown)


mcalmer (approved):
LGTM

PR #1055, opened 2024-01-04
Set the correct base config module for salt-migration-minion

NamelessOne91:
Title - It was pointing to a non-existent "base" module

### DIFF
--- a/terracumber_config/tf_files/SUSEManager-4.3-build-validation-NUE.tf
+++ b/terracumber_config/tf_files/SUSEManager-4.3-build-validation-NUE.tf
@@ -805,12 +805,12 @@ module "sles15sp5s390-minion" {
 }
 
 module "salt-migration-minion" {
-  source = "./modules/minion"
-  base_configuration = module.base.configuration
+  source             = "./modules/minion"
+  base_configuration = module.base_core.configuration
   name               = "min-salt-migration"
   product_version    = "4.3-released"
   image              = "sles15sp5o"
-  provider_settings = {
+  provider_settings  = {
     mac                = "aa:b2:92:42:00:cf"
     memory             = 4096
   }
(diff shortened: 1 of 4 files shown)


terracumber_config/tf_files/Uyuni-Master-build-validation-NUE.tf
```diff
@@ -638,8 +638,8 @@ module "sles15sp5s390-minion" {
 }
 
 module "salt-migration-minion" {
-  source = "./modules/minion"
-  base_configuration = module.base.configuration
+  source             = "./modules/minion"
+  base_configuration = module.base_core.configuration
```
srbarrios, line 642: I would better use the same base for all the terraform files, `base_new_sle` is fine.
But please, avoid using `base_core`, having this setup, it will deploy this VM together with the server and core components, and we want to keep core components as much isolated as possible.

terracumber_config/tf_files/SUSEManager-4.3-build-validation-NUE.tf
```diff
@@ -805,12 +805,12 @@ module "sles15sp5s390-minion" {
 }
 
 module "salt-migration-minion" {
-  source = "./modules/minion"
-  base_configuration = module.base.configuration
+  source             = "./modules/minion"
+  base_configuration = module.base_core.configuration
```
srbarrios, line 809: let's use `base_new_sle`

terracumber_config/tf_files/Uyuni-Master-build-validation-NUE.tf
Bischoff: what you say is true for PRV, not for NUE, where we have only one hypervisor

terracumber_config/tf_files/SUSEManager-4.3-build-validation-NUE.tf
Bischoff: same

Bischoff (approved):
still a red, i will fix it separately

PR #1057, opened 2024-01-04
Let sumaform tests pass with the s390 provider

Bischoff:
One design problem of current sumaform is that we can have only one backend symlink for all modules.

The sumaform tests on github actions set this symlink to the null provider, but the s390 module still points to the s390 provider for its backend.

This PR solves the issue by installing the s390 provider and tweaking the tf files a bit more.


maximenoel8 (approved):
LGTM

srbarrios (approved):

PR #1063, opened 2024-01-08
Keep database content inside the root partition on BV

srbarrios:
Keep database content inside the root partition on BV.
Otherwise we have too many free space in the root partition that it will never be used.

Note: We might have a similar consequence in our CI envs, but for now, we will keep it as it is there, as we are unsure about how to resize the root partition and the current size (200GB) is not enough for Database Backup/Restore test.


Bischoff (approved):
Please note that space is not really a problem, as qcow2 disks grow on demand... but yeah a simplification is always welcome.

This will also align 4.3 bv with uyuni BV.

PR #1065, opened 2024-01-09
Increase repository_disk_size on Uyuni CI, as we synchronize two products SLES15SP4 and Leap 15.5

srbarrios:
Increase repository_disk_size on Uyuni CI, as we synchronize two products SLES15SP4 and Leap 15.5


Bischoff (approved):
Thank you

PR #1069, opened 2024-01-11
Adding Salt Shaker environments and TF files

meaksh:
This PR just adds the environments and TF files needed to run the Salt Shaker jobs in our Jenkins: https://ci.suse.de/user/manager/my-views/view/Salt%20Shaker/

Interesting bits:
- We have a total of 52 new environments + each correpsonding TF files.
- New pipeline used for those jobs: `pipeline-salt-shaker.groovy`
- The virthost and pipeline schedule are set according to plan defined at https://github.com/SUSE/spacewalk/issues/23286#issuecomment-1876792916
- All Salt Shaker environments and TF files located in dedicated directories.


Bischoff (approved):

ktsamis (approved):
Overall looks good, I can't review all the files but they are similar for different envs. If they were running before they should run now.

PR #1077, opened 2024-01-15
Make sure URLTrigger finds nodes to run for Salt Shaker

meaksh:
I'm having issues to make "URLTrigger" plugin to find nodes where to run.

According to https://www.jenkins.io/doc/pipeline/steps/params/pipelinetriggers/ , now passing both `triggerLabel` and `labelRestriction` parameters I think this should work.


m-czernek (approved):
LGTM

PR #1078, opened 2024-01-16
Fix issue on how parameters are passed to URLTrigger

meaksh:
I just realized there was a problem on how the parameters were passed to `URLTrigger` and that is what was causing the final pipeline didn't get the expected configuration for URLTrigger.

Reference: "URLTrigger section" at https://www.jenkins.io/doc/pipeline/steps/params/pipelinetriggers/


m-czernek (approved):
:+1:

PR #1079, opened 2024-01-18
Enable catch timeout message for BV pipelines

maximenoel8:
## What does this PR 

Use the new controller variable created in https://github.com/uyuni-project/sumaform/pull/1465 to enable the catch timeout mechanism during BV pipelines.


srbarrios:
We might be missing something, check the GH check:
```
╷
│ Error: Unsupported argument
│ 
│   on main.tf line 1362, in module "controller":
│ 1362:   catch_timeout_message = true
│ 
│ An argument named "catch_timeout_message" is not expected here.
╵
```

srbarrios (approved):
LGTM. But first, let's merge the sumaform PR, and re-trigger the GH check to have it green.

PR #1098, opened 2024-01-29
Add s390x minion and remove opensuse153arm minion

ktsamis:
Add s390x minion and remove opensuse153arm minion and opensuse154arm minion, they are both EOL


Bischoff:
I did not think about the grace period, we might have to partially revert it

PR #1106, opened 2024-01-31
Add 4.3 VM pipeline

jordimassaguerpla:
https://github.com/SUSE/spacewalk/issues/22625


jordimassaguerpla:
@Bischoff : Can I use suma-01.mgr.suse.de as for the hypervisor? Or should I look for another hypervisor?

jordimassaguerpla:
Hi! With Eric we decided to use suma08 for the suma 4.3 new vm pipeline. This is temporary until we pass all the tests and then we will replace the suma 4.3 pipeline with it

Issue #1108, opened 2024-02-05
BV pipeline is green even when the proxy fails to bootstrap


NamelessOne91:
The issue has been noticed on the 4.3.11 BV NUE pipeline.

Running a build with these option flagged

- enable_proxy_stages
- must_boot_node

Should produce a failure at the pipeline level in case one of the proxy stages fails.

Right now it does not and the pipeline is shown as "green".

Take [Build 167](https://ci.suse.de/view/Manager/view/Manager-4.3/job/manager-4.3-qe-build-validation-NUE/167/) for reference

NamelessOne91:
Docs for the sh global function state this about using **returnStatus**

> Normally, a script which exits with a nonzero status code will cause the step to fail with an exception. If this option is checked, the return value of the step will instead be the status code. You may then compare it to zero, for example

Currently most of the steps for the proxy save the `returnStatus` to a variable but don't use it to determine whether a failure occurred or not. The same holds true for the monitoring server.
Only the MUs stages will properly fail if a script execution does not return with 0.

nodeg:
I can confirm this issue. I saw the same with the monitoring stages.


![Image](https://github.com/SUSE/susemanager-ci/assets/12104291/d782f012-fc46-45b9-ba55-81297c764275)

![Image](https://github.com/SUSE/susemanager-ci/assets/12104291/503390a0-d534-4b7a-b4a0-61040d893396)

![Image](https://github.com/SUSE/susemanager-ci/assets/12104291/782a03d4-a577-4a0d-a3b2-8d4b541912bd)

NamelessOne91:
Discussed approaches

1) Fail fast: if a test fails, the pipeline fails.

This mimics whats a lot of us do when they are actually working on BV, as they would likely stop and debug the first failure that occurs. It saves time and avoids having misleading "greens" being shown in the UI and reports.

2) Fail only in critical stages

The pipeline should fail only when a stage that is crucial for the following ones fails. In other cases the pipeline keeps on running and the errors are stored to be reported at the end.
This introduces a degree of complexity but allows to leave the pipeline running extensive parts of the testsuite and check the results later on (example: runs over the weekend).

3) Fine grained control over failures

Use a variable to conditionally decide whether to fail fast or not. 
More complexity VS More control  kind of situation.

nodeg:
I just saw, that this issue was opened in `susemanager-ci`. If you create an issue, please always open it in `spacewalk`, so we have all our labels, etc. available.

PR #1110, opened 2024-02-06
BV: fail-fast on non-parallel stages' failures

NamelessOne91:
Closes https://github.com/SUSE/susemanager-ci/issues/1108

Docs for the sh global function state this about using returnStatus

>  Normally, a script which exits with a nonzero status code will cause the step to fail with an exception. If this option is checked, the return value of the step will instead be the status code. You may then compare it to zero, for example

Currently most of the steps for the proxy save the returnStatus to a variable but don't use it to determine whether a failure occurred or not. The same holds true for the monitoring server.
Only the MUs stages will properly fail if a script execution does not return with 0.

### DIFF
--- a/jenkins_pipelines/environments/common/pipeline-build-validation.groovy
+++ b/jenkins_pipelines/environments/common/pipeline-build-validation.groovy
@@ -13,11 +13,6 @@ def run(params) {
 
         // Declare lock resource use during node bootstrap
         mgrCreateBootstrapRepo = 'share resource to avoid running mgr create bootstrap repo in parallel'
-        // Variables to store none critical stage run status
-        def monitoring_stage_result_fail = false
-        def client_stage_result_fail = false
-        def retail_stage_result_fail = false
-        def containerization_stage_result_fail = false
 
         env.common_params = "--outputdir ${resultdir} --tf ${params.tf_file} --gitfolder ${resultdir}/sumaform"
 
@@ -44,8 +39,9 @@ def run(params) {
                 }
             }
 
-            stage('Deploy') {
-                if (params.must_deploy) {
+            /** Deploy stage begin **/
+            if(params.must_deploy) {
+                stage('Deploy') {
                     // Generate json file in the workspace
                     if (params.custom_repositories?.trim()) {
                         // JSON file passed by parameter
@@ -74,19 +70,23 @@ def run(params) {
                     deployed = true
                 }
             }
+            /** Deploy stage end **/
 
+            /** Sanity check stage - always run **/
             stage('Sanity check') {
                 sh "./terracumber-cli ${common_params} --logfile ${resultdirbuild}/testsuite.log --runstep cucumber --cucumber-cmd 'cd /root/spacewalk/testsuite; ${env.exports} rake cucumber:build_validation_sanity_check'"
             }
 
-            stage('Run core features') {
-                if (params.must_run_core && (deployed || !params.must_deploy)) {
+            /** Core features stage **/
+            if (params.must_run_core && (deployed || !params.must_deploy)) {
+                stage('Run core features') {
                     sh "./terracumber-cli ${common_params} --logfile ${resultdirbuild}/testsuite.log --runstep cucumber --cucumber-cmd '${env.exports} cd /root/spacewalk/testsuite; rake cucumber:build_validation_core'"
                 }
             }
-
-            stage('Sync. products and channels') {
-                if (params.must_sync && (deployed || !params.must_deploy)) {
+            
+            /** Reposync stage **/
+            if (params.must_sync && (deployed || !params.must_deploy)){
+                stage('Sync. products and channels') {
                     // Get minion list from terraform state list command
                     def nodesHandler = getNodesHandler()
                     res_products = sh(script: "./terracumber-cli ${common_params} --logfile ${resultdirbuild}/testsuite.log --runstep cucumber --cucumber-cmd 'unset ${nodesHandler.envVariableListToDisable.join(' ')}; ${env.exports} cd /root/spacewalk/testsuite; rake cucumber:build_validation_reposync'", returnStatus: true)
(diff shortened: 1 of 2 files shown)


Bischoff (approved):
please don't merge until Maxime had a look

ktsamis (approved):
I remember fixing something similar in normal BV only but that's a good idea to do everywhere. The exit code is anyway taken into account in jenkins as a failed step and it fails the stage but why not be explicit everywhere? This makes sense to me

srbarrios:
To run as much as we can of our Build Validation, we intentionally decided to don't stop the pipeline in some stages. As an example, if monitoring stage fails, that's not a reason to stop run the boostrap of all our minions. In theory, we have stored the result and we expose it at the end of the pipeline.
But I would agree to fail the whole pipeline if all the team agree on that way of proceeding.

In any case, if we decide to fail in one of those stages changed in this PR, then we better also refactor some code like:

https://github.com/SUSE/susemanager-ci/blob/master/jenkins_pipelines/environments/common/pipeline-build-validation.groovy#L17-L20
https://github.com/SUSE/susemanager-ci/blob/master/jenkins_pipelines/environments/common/pipeline-build-validation.groovy#L187-L190
https://github.com/SUSE/susemanager-ci/blob/master/jenkins_pipelines/environments/common/pipeline-build-validation.groovy#L292-L303

NamelessOne91:
Refactored & Rebased

The "standard" BV pipeline should now fail whenever a stage fails.
The only exception is when a failure occurs in a stage that is being run in parallel ( as of now: clients testing):  the interested parallel "branch" should be marked as failing and terminate but the pipeline should continue running the other branches and the following stages.

I also did a minor refactor around the checks on pipeline params like  `must_create_X` and `enable_X' in order to not display at all stages which are  not running any test.

I'll push my local changes to the AWS BV pipeline once we agree on these ones and we decide to have the same (or another) behavior there.

ktsamis (approved):

maximenoel8:
|| I also did a minor refactor around the checks on pipeline params like must_create_X and enable_X in order to not create and display at all stages which are not running any test.

We are keeping all the stages even if they are skip for a better vision in the main jenkins view ( when you open the build ).
If the number of stages change, it will wipe the other stages from the view.

It was an explicit demands from @Bischoff

maximenoel8:
I'm against reverting the stages change. We had a long discussion about it and we decided to keep the skip stages visible.

maximenoel8:
I'm also against removing the try / catch mechanism. This will stop the pipeline when we don't want. Having a failing stage is putting the stage in red and at the end if any stages failed the pipeline is red.

Let's organize a meeting to understand the need of this full refactor.

We reach this point by iteration and those changes are removing lot of features that were added later after long discussion with the team.

maximenoel8 (changes requested):
I don't really understand the need of those changes and it reverts the improvements added during the use of the BV pipeline by the team.

jenkins_pipelines/environments/common/pipeline-build-validation.groovy
```diff
@@ -74,181 +70,184 @@ def run(params) {
...
-
-            stage('Sync. products and channels') {
-                if (params.must_sync && (deployed || !params.must_deploy)) {
+            
+            /** Reposync stage **/
+            if (params.must_sync && (deployed || !params.must_deploy)){
+                stage('Sync. products and channels') {
                     // Get minion list from terraform state list command
                     def nodesHandler = getNodesHandler()
                     res_products = sh(script: "./terracumber-cli ${common_params} --logfile ${resultdirbuild}/testsuite.log --runstep cucumber --cucumber-cmd 'unset ${nodesHandler.envVariableListToDisable.join(' ')}; ${env.exports} cd /root/spacewalk/testsuite; rake cucumber:build_validation_reposync'", returnStatus: true)
```
maximenoel8, line 92: If the issue is related to reposync failing, we probably only need to change this for : 
sh "./terracumber-cli ${common_params} --logfile ${resultdirbuild}/testsuite.log --runstep cucumber --cucumber-cmd 'unset ${nodesHandler.envVariableListToDisable.join(' ')}; ${env.exports} cd /root/spacewalk/testsuite; rake cucumber:build_validation_reposync'"

maximenoel8:
Did you test your changes ? Tried the different scenarios ?

srbarrios:
I agree with @maximenoel8 points of view.

NamelessOne91:
> I don't really understand the need of those changes and it reverts the improvements added during the use of the BV pipeline by the team.

To give some context, the actual issues we had are the ones mentioned in the first comment: some proxy (and monitoring) stages where still marked as green even when failing. 
That's due to us not checking the status code returned by sh commands in those stages, I don't know whether that's intentional or just something that slipped through in the past.
Anyway, the PR was limited to that scope initially.

After discussing it, there has been a request to see if could make the pipeline fail-fast as that would have been closer to the workflow some of us adopt anyway.

> I'm against reverting the stages change. We had long discussions about it and we decided to keep the skip stages visible.

That's actually something I did on my initiative, so I'm totally fine reverting my changes and keep stages that are skipped still visible. I can see the point in doing so.

All that said, it'll be perfectly fine for me to freeze the refactor and/or discard it entirely.
I think the only issue that actually might have to be addressed is the one mentioned in the 1st comment .

srbarrios:
FMPOV, the only remarkable improvement is to fail the full pipeline if the proxy fails, as it's a strong dependency for all the onboarding of clients. But I don't know if we missed this behavior in our refactor or it was made in purpose.

Let's have a quick discussion tomorrow after the planning meeting?

maximenoel8:
I was playing around today with the pipeline, I think we are missing an error for those stages : 
 - Add Activation Keys Proxy
 - Create bootstrap repository Proxy
 - Bootstrap Proxy
 - Add Activation Keys Monitoring
 - Create bootstrap repository Monitoring
 - Bootstrap Monitoring Server
Adding a step for those stages either using : 
```
                    if (res_XXXX != 0) {
                        error("Stage fails ${res_XXXX}")
                    }
```

or 

```
                    sh "exit ${res_XXX}"
```

should do the trick.

NamelessOne91:
Closing as we agreed to only add the proxy stages failures and make the pipeline fail if they fail.

PR #1115, opened 2024-02-09
Add the Salt bundle to all minions in the 5.0 BV

nodeg:
According to https://github.com/uyuni-project/sumaform/blob/master/README_TESTING.md#using-salt-bundle-venv-salt-minion-in-head-and-uyuni, we should use the Salt bundle for all minions on HEAD. Pablo confirmed this in [Slack](https://suse.slack.com/archives/C02CY2CLLH3/p1707484962948199).

I am not 100 % sure if this will work for the SLE Micro 5.x minions and the new s390 ones.


ktsamis (approved):

vandabarata (approved):

NamelessOne91 (approved):

nodeg:
Agreed. But that is something for the future, when we only use the Salt bundle everywhere.

PR #1117, opened 2024-02-13
Add support for 5.0 in json script

ktsamis:
Added Alma 8 client tools.

Added 5.0 version in command line. For now 4.3 is still default.

Added 5.0 beta client tools. They will be separate from 4.3. They won't become common at any point even after GA. Until the betas are done we need 5.0 beta and non-beta client tools in the minions, they will get non-beta from the testsuite anyway but wanted to add the possibility of a parallel MI that has beta and non-beta, especially during 4.3 MU while a beta is running.

Added hardcoded 5.0 server and proxy repo. Unfortunately for this we won't be getting MIs but only directly updating this url with the latest. 

Added possibility for mi_ids to be separated by spaces as well as comma. 

I know it's a bit hacky and long now but a refactor has to wait a bit until the necessary functions work and I have time


jenkins_pipelines/scripts/maintenance_json_generator.py
```diff
@@ -44,11 +44,12 @@
...
-                             "/SUSE_Updates_SLE-Module-Server-Applications_15-SP5_s390x/"],
+                            "/SUSE_Updates_SLE-Module-Basesystem_15-SP5_s390x/",
+                            "/SUSE_Updates_SLE-Module-Server-Applications_15-SP5_s390x/"],
     "centos7_client": "/SUSE_Updates_RES_7-CLIENT-TOOLS_x86_64/",
     "centos7_minion": "/SUSE_Updates_RES_7-CLIENT-TOOLS_x86_64",
     "rocky8_minion": "/SUSE_Updates_RES_8-CLIENT-TOOLS_x86_64/",
+    "alma8_minion": "/SUSE_Updates_RES_8-CLIENT-TOOLS_x86_64/",
     "ubuntu2004_minion": "/SUSE_Updates_Ubuntu_20.04-CLIENT-TOOLS_x86_64/",
     "ubuntu2204_minion": "/SUSE_Updates_Ubuntu_22.04-CLIENT-TOOLS_x86_64/",
     "debian10_minion": "/SUSE_Updates_Debian_10-CLIENT-TOOLS_x86_64/",
```
deneb-alpha, line 55: Debain10 is gone 
```suggestion
```

jenkins_pipelines/scripts/maintenance_json_generator.py
```diff
@@ -80,7 +81,52 @@
...
+    "sle15sp5_client": "/SUSE_Updates_SLE-Manager-Tools_15-BETA_x86_64/",
+    "sle15sp5_minion": "/SUSE_Updates_SLE-Manager-Tools_15-BETA_x86_64/",
+    "sle15sp5s390_minion": "/SUSE_Updates_SLE-Manager-Tools_15-BETA_s390x/",
+    "centos7_client": "/SUSE_Updates_RES_7-CLIENT-TOOLS-BETA_x86_64/",
+    "centos7_minion": "/SUSE_Updates_RES_7-CLIENT-TOOLS-BETA_x86_64",
+    "rocky8_minion": "/SUSE_Updates_RES_8-CLIENT-TOOLS-BETA_x86_64/",
+    "alma8_minion": "/SUSE_Updates_RES_8-CLIENT-TOOLS-BETA_x86_64/",
+    "ubuntu2004_minion": "/SUSE_Updates_Ubuntu_20.04-CLIENT-TOOLS-BETA_x86_64/",
+    "ubuntu2204_minion": "/SUSE_Updates_Ubuntu_22.04-CLIENT-TOOLS-BETA_x86_64/",
+    "debian10_minion": "/SUSE_Updates_Debian_10-CLIENT-TOOLS-BETA_x86_64/",
```
deneb-alpha, line 107: Debian 10 is EOL
```suggestion
```

deneb-alpha (changes requested):
First round of review from my side.  

Until we split the client tools between 4.3 and 5.0 I  think the script should be ok.  
in any case, I left a few suggestions

jenkins_pipelines/scripts/maintenance_json_generator.py
ktsamis: Removed, thanks Marina

jenkins_pipelines/scripts/maintenance_json_generator.py
ktsamis: Removed, thanks :)

Bischoff (approved):
approving but i'm unable to really judge the details

deneb-alpha (approved):

PR #1125, opened 2024-02-21
Remove Salt Bundle from SLE Micro SSH Minions

nodeg:
This PR

- fixes the CI with respect to the Feilong version bump (0.0.4 -> 0.0.6)
- removes the Salt bundle for SLE Micro SSH Minions (This is not supported, yet. See https://bugzilla.suse.com/show_bug.cgi?id=1208045)

### DIFF
--- a/terracumber_config/tf_files/SUSEManager-5.0-build-validation-NUE.tf
+++ b/terracumber_config/tf_files/SUSEManager-5.0-build-validation-NUE.tf
@@ -1146,8 +1146,9 @@ module "slemicro51-sshminion" {
   use_os_released_updates = false
   ssh_key_path            = "./salt/controller/id_rsa.pub"
 
-  additional_packages = [ "venv-salt-minion" ]
-  install_salt_bundle = true
+//  WORKAROUND until https://bugzilla.suse.com/show_bug.cgi?id=1208045 gets fixed  
+//  additional_packages = [ "venv-salt-minion" ]
+//  install_salt_bundle = true
 }
 
 module "slemicro52-sshminion" {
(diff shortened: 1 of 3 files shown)


Bischoff (commented):
thanks for the feilong catch

as this is a bug, rather than removing, should we comment out instead, with a reference to the bsc?

terracumber_config/tf_files/SUSEManager-5.0-build-validation-NUE.tf
```diff
@@ -1146,8 +1146,9 @@ module "slemicro51-sshminion" {
   use_os_released_updates = false
   ssh_key_path            = "./salt/controller/id_rsa.pub"
 
-  additional_packages = [ "venv-salt-minion" ]
-  install_salt_bundle = true
+//  WORKAROUND until https://bugzilla.suse.com/show_bug.cgi?id=1208045 gets fixed  
+//  additional_packages = [ "venv-salt-minion" ]
+//  install_salt_bundle = true
```
srbarrios, line 1151: I wonder if we should change sumaform, so `install_salt_bundle` has as default a true value. This will align our sumaform deployments with the official doc for 4.3 and 5.0

About `additional_packages = [ "venv-salt-minion" ]` I wonder if we really need this here, in theory, in the cloud-init phase we are also forcing to install this package for almost all the Operating Systems (very few keep using the salt-minion package)

srbarrios (approved):
Approved with a comment, as maybe we should think on improving sumaform code, instead of keeping this commented due to a bug.

terracumber_config/tf_files/SUSEManager-5.0-build-validation-NUE.tf
Bischoff: `install_salt_bundle` having by default a true value -> absolutely

and later on we could even remove these 2 variable and completely forget about regular salt-minion

PR #1128, opened 2024-02-22
BV: Use the correct 5.0 Beta 1 URL

nodeg:
This updates the server container URL for the 5.0 Beta 1 BV.


ktsamis (approved):
Looks correct to me, if Marina gives the ok, go ahead :+1:

NamelessOne91 (approved):

deneb-alpha (approved):
I'll try to simplify the number of repos we use for containers. 

For now it's ok to use containerfile, as soon as I'll be able to just use one I'll ping you back for adjusting this path.

In any case, the repo will stay the same for beta1

PR #1138, opened 2024-03-04
BV:  allow a conditional check for MIs under embargo

NamelessOne91:
Closes https://github.com/SUSE/spacewalk/issues/21702
Partially related to https://github.com/SUSE/spacewalk/issues/22533  and https://github.com/SUSE/spacewalk/issues/23559 (unit tests, improvements)

Adds an additional flag to the `maintenance_json_generator.py` script  allowing to check if each passed MI is (still) under embargo and exclude it from being considered in the generated JSON file.

This is done by using the `osc` CLI to query each MI project attributes and patchinfo file + SMASH API to verify which issues are still considered under embargo.
Buildops mentioned that MIs containing fixes under embargo should have the attribute `OBS:EmbargoDate` set to the embargo end.

Includes a refactor of the script to:

- improve the CLI args parsing
- use sets when possible, to avoid having duplicate IDs and repositories
- cache calls to the create_url function, in order to reduce HTTP calls
- add type hints and make the code a bit more DRY
- have separate files and classes to handle calls to external APIs
- add unit tests

You can run tests positioning yourself in the `scripts`  folder and executing the following command:

`python3 -m unittest discover`


jenkins_pipelines/scripts/maintenance_json_generator.py
```diff
@@ -2,139 +2,143 @@
 import json
 import subprocess
 import sys
-
 import requests
+import xml.etree.ElementTree as ET
+from datetime import date, datetime
+from functools import cache
```
szachovy, line 8: Please follow the recommended import order https://peps.python.org/pep-0008/#imports

jenkins_pipelines/scripts/maintenance_json_generator.py
```diff
@@ -2,139 +2,143 @@
...
-    "sle15sp1_client": ["/SUSE_Updates_SLE-Manager-Tools_15_x86_64/",
-                        "/SUSE_Updates_SLE-Product-SLES_15-SP1-LTSS_x86_64/"],
-    "sle15sp1_minion": ["/SUSE_Updates_SLE-Manager-Tools_15_x86_64/",
-                        "/SUSE_Updates_SLE-Product-SLES_15-SP1-LTSS_x86_64/"],
-    "sle15sp2_client": ["/SUSE_Updates_SLE-Manager-Tools_15_x86_64/",
-                        "/SUSE_Updates_SLE-Product-SLES_15-SP2-LTSS_x86_64/"],
-    "sle15sp2_minion": ["/SUSE_Updates_SLE-Manager-Tools_15_x86_64/",
-                        "/SUSE_Updates_SLE-Product-SLES_15-SP2-LTSS_x86_64/"],
-    "sle15sp3_client": ["/SUSE_Updates_SLE-Manager-Tools_15_x86_64/",
+clienttools43dict: dict[str, set[str]] = {
```
szachovy, line 13: Good practice is to do not express the type inside the variable name e.g. `my_list: list`, so I would just go for:
```suggestion
clienttools43: dict[str, set[str]] = {
```

The same with other variable names covered by this PR.

jenkins_pipelines/scripts/maintenance_json_generator.py
```diff
@@ -2,139 +2,143 @@
...
-    "slemicro52_minion": ["/SUSE_Updates_SLE-Manager-Tools-BETA-For-Micro_5_x86_64/",
-                          "/SUSE_Updates_SLE-Manager-Tools_15-BETA_x86_64/"],
-    "slemicro53_minion": ["/SUSE_Updates_SLE-Manager-Tools-BETA-For-Micro_5_x86_64/",
-                          "/SUSE_Updates_SLE-Manager-Tools_15-BETA_x86_64/"],
-    "slemicro54_minion": ["/SUSE_Updates_SLE-Manager-Tools-BETA-For-Micro_5_x86_64/",
-                          "/SUSE_Updates_SLE-Manager-Tools_15-BETA_x86_64/"],
-    "slemicro55_minion": ["/SUSE_Updates_SLE-Manager-Tools-BETA-For-Micro_5_x86_64/",
-                          "/SUSE_Updates_SLE-Manager-Tools_15-BETA_x86_64/"],
-    "salt_migration_minion": "/SUSE_Updates_SLE-Manager-Tools_15_x86_64-BETA/"
+clienttools50betadict: dict[str, set[int]] = {
```
szachovy, line 88: I think you meant
```suggestion
clienttools50betadict: dict[str, set[str]] = {
```

jenkins_pipelines/scripts/maintenance_json_generator.py
```diff
@@ -148,174 +152,160 @@
...
-                finaldict[key] = value + [clienttools50betadict[key]]
-        else:
-            if isinstance(clienttools50betadict[key], list):
-                finaldict[key] = [value] + clienttools50betadict[key]
-            else:
-                finaldict[key] = [value, clienttools50betadict[key]]
-    else:
-        finaldict[key] = value
-
+        finaldict[key].update(clienttools50betadict[key])
```
szachovy, line 172: You can consider using something more pythonic like:
```
finaldict: dict[str, set[str]] = {key: value.union(clienttools50betadict.get(key, set())) for key, value in clienttools43dict.items()}
```
similarly below.

jenkins_pipelines/scripts/maintenance_json_generator.py
```diff
@@ -148,174 +152,160 @@
...
+    return output
+    
+def read_requests() -> set[str]:
+    # Find open requests
+    # TODO Find a better way to query the open requests, this is fragile because it depends on external utils
+    # being there.
+    output: str = osc_ibs_request("qam open -G qam-manager")
+    lines: list[str] = output.splitlines()
+    # Create a set of the maintenance incidents from the output
+    mi_ids = { line.rstrip().split(sep=":")[3] for line in lines if "ReviewRequest" in line }
```
szachovy, line 219: If it's fragile I would try to catch the eventual exceptions what this function may raise if external units are not readable.

jenkins_pipelines/scripts/maintenance_json_generator.py
```diff
@@ -148,174 +152,160 @@
...
     return mi_ids
 
+@cache
+def create_url(mi_id:str, suffix: str) -> str:
```
szachovy, line 223: Code should either have all the returns or no returns at all, thus:
```suggestion
def create_url(mi_id:str, suffix: str) -> str | None:
```

jenkins_pipelines/scripts/maintenance_json_generator.py
```diff
@@ -148,174 +152,160 @@
...
-def find_valid_repos(mi_ids, version):
-    if version == '43':
-        dict_version = nodesdict43
-    elif version == '50':
-        dict_version = nodesdict50
-    else:
-        print("You have not given one of the correct options, run the script with -h to see the correct ones")
+    res = requests.get(url)
+    if res.ok:
+        return url
```
szachovy, line 228: and
```suggestion
    if res.ok:
        return url
    return None
```

jenkins_pipelines/scripts/maintenance_json_generator.py
```diff
@@ -148,174 +152,160 @@
...
+                        if final_id in custom_repositories[node]:
+                            for i in range(1, 100):
+                                new_id = f"{mi_id}-{i}" 
+                                if new_id not in custom_repositories[node]:
+                                    final_id = new_id
+                                    break
+                        # for each mi_id we have multiple repos sometimes for each node
+                        custom_repositories[node][final_id] = repo
+                    else:
+                        custom_repositories[node] = {mi_id: repo}
```
szachovy, line 264: This code has like infinite time complexity, but it's for a separate PR.

jenkins_pipelines/scripts/maintenance_json_generator.py
```diff
@@ -148,174 +152,160 @@
...
-        return url
-
+    embargo_attribute = xml_attributes.find("./attribute[@name='EmbargoDate'][value]")
+    if not embargo_attribute:
+        print(f"MI {mi_id} is not under embargo")
+        return False
+    
+    # TODO: verify what's the actual date format used in the EmbargoDate attribute
+    embargo_end_date_str = embargo_attribute.find("./value").text
+    embargo_end_date = datetime.strptime(embargo_end_date_str, "%Y-%m-%d").date()
```
szachovy, line 288: Types in this variables would also help.

jenkins_pipelines/scripts/maintenance_json_generator.py
```diff
@@ -148,174 +152,160 @@
...
+    if embargo_end_date >= CURRENT_DATE:
+        print(f"MI {mi_id} is under embargo until {embargo_end_date}. Today is {CURRENT_DATE}")
+        return True
+    
+    print(f"MI {mi_id} embargo was lifted after {embargo_end_date}. Today is {CURRENT_DATE}")
+    return False
```
szachovy, line 294:
```suggestion
    embargo_attribute = xml_attributes.find("./attribute[@name='EmbargoDate'][value]")
    if embargo_attribute:    
        # TODO: verify what's the actual date format used in the EmbargoDate attribute
        embargo_end_date_str = embargo_attribute.find("./value").text
        embargo_end_date = datetime.strptime(embargo_end_date_str, "%Y-%m-%d").date()
        if embargo_end_date >= CURRENT_DATE:
            print(f"MI {mi_id} is under embargo until {embargo_end_date}. Today is {CURRENT_DATE}")
            return True
        else:
            print(f"MI {mi_id} embargo was lifted after {embargo_end_date}. Today is {CURRENT_DATE}")
    else:
        print(f"MI {mi_id} is not under embargo")
    return False
```

jenkins_pipelines/scripts/maintenance_json_generator.py
```diff
@@ -148,174 +152,160 @@
...
         # Ignore spaces before or after the comma
-        mi_ids = [id.strip() for id in args.mi_ids.split(",")]
+        mi_ids: set[str] = { id.strip() for id in args.mi_ids.split(",") }
     else:
-        mi_ids = read_requests()
-    find_valid_repos(mi_ids, args.version)
+        mi_ids: set[str] = read_requests()
 
+    if args.embargo_check:
+        mi_ids: set[str] = { id for id in mi_ids if not is_under_embargo(id) }
```
szachovy, line 306: It's not good to reuse the same variable name with a different content in the same function.

jenkins_pipelines/scripts/maintenance_json_generator.py
NamelessOne91: Right, I kept the previous names but they do not make make sense once type hints are in place.

jenkins_pipelines/scripts/maintenance_json_generator.py
NamelessOne91: Indeed, thanks for the catch !

jenkins_pipelines/scripts/maintenance_json_generator.py
NamelessOne91: I'm fine replacing it with a comprehension but others may find it more readable as its is.
I'll check later for comments on this.

jenkins_pipelines/scripts/maintenance_json_generator.py
NamelessOne91: `osc_ibs_request` has some internal error handling which can be moved up the stack and/or improved.

jenkins_pipelines/scripts/maintenance_json_generator.py
NamelessOne91: Good catch.
I'd probably return an empty string instead of `None` though.
Not that it matters that much for the context of this script , as they both would evaluate to False where the return is used.

jenkins_pipelines/scripts/maintenance_json_generator.py
NamelessOne91: I see the point, but I'm not sure in this specific case it would  be better to have something like

```python
if args.embargo_check:
        no_embargo_mi_ids: set[str] = { id for id in mi_ids if not is_under_embargo(id) }
        find_valid_repos(no_embargo_mi_ids, args.version)
else:
        find_valid_repos(mi_ids, args.version)
```

just to avoid reassigning the variable, given that it'll not be used for anything else after the function call

srbarrios (approved):
Super nice job @NamelessOne91 
LGTM thank you.

NamelessOne91:
A part from adding more unit tests and addressing future reviews, there's 3 things left to discuss from my POV:

1) Decide what to do when an MI is found to be under embargo. Currently the script discards it from the final JSON and keeps running, but we may want to fail instead. Or use another optional flag to determine what to do.

2) When no MIs are passed to the scripts, it calls the following command that does not work for me on local (osc V 1.6.1):
  
  ```bash
  paolo@linux:~/SUSE/susemanager-ci/jenkins_pipelines/scripts> osc -A https://api.suse.de qam open -G qam-manager
usage: osc [global opts] <command> [--help] [opts] [args]
osc: error: argument command: invalid choice: 'qam'
  ```
  
 3) Double check how the script is used in the BV pipeline and use it in the Cloud ones with the new flag enabled - likely to be material for another PR

It was present in the original script  under a comment saying it is fragile because it depends on external tools. Is it expected to work only in BV ?

maximenoel8 (approved):
LGTM

NamelessOne91:
>     3. Double check how the script is used in the BV pipeline and use it in the Cloud ones with the new flag enabled - likely to be material for another PR.

This PR should fix the usage of the script's output in standard BV pipelines.
Adding the same behaviour, with the embargo check enabled, to the 4.3 BV AWS pipeline will be handled by https://github.com/SUSE/susemanager-ci/pull/1152

vandabarata (approved):

srbarrios (approved):
LGTM

PR #1139, opened 2024-03-06
BV: check sh return code in all proxy and monitoring stages

NamelessOne91:
Closes https://github.com/SUSE/susemanager-ci/issues/1108

Docs for the sh global function state this about using returnStatus

>Normally, a script which exits with a nonzero status code will cause the step to fail with an exception. If this option is checked, the return value of the step will instead be the status code. You may then compare it to zero, for example

Currently most of the steps for the proxy save the returnStatus to a variable but don't use it to determine whether a failure occurred or not. The same holds true for the monitoring server.
Only the MUs stages will properly fail if a script execution does not return with 0.

### DIFF
--- a/jenkins_pipelines/environments/common/pipeline-build-validation-aws.groovy
+++ b/jenkins_pipelines/environments/common/pipeline-build-validation-aws.groovy
@@ -262,6 +262,7 @@ def run(params) {
                     }
                     res_add_keys = sh(script: "./terracumber-cli ${common_params} --logfile ${resultdirbuild}/testsuite.log --runstep cucumber --cucumber-cmd '${env.exports} cd /root/spacewalk/testsuite; rake cucumber:build_validation_add_activation_key_proxy'", returnStatus: true)
                     echo "Add Proxy Activation Key status code: ${res_add_keys}"
+                    sh "exit ${res_add_keys}"
                 }
             }
             stage('Create bootstrap repository Proxy') {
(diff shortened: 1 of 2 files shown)


jenkins_pipelines/environments/common/pipeline-build-validation-aws.groovy
```diff
@@ -352,6 +357,7 @@ def run(params) {
                             echo 'Register monitoring server as minion with gui'
                             res_init_monitoring = sh(script: "./terracumber-cli ${common_params} --logfile ${resultdirbuild}/testsuite.log --runstep cucumber --cucumber-cmd '${env.exports} cd /root/spacewalk/testsuite; rake cucumber:build_validation_init_monitoring'", returnStatus: true)
                             echo "Init Monitoring Server status code: ${res_init_monitoring}"
+                            sh "exit ${res_init_monitoring}"
```
srbarrios, line 360: I would only do this change for proxy, not for monitoring. We can keep testing without monitoring.

srbarrios (approved):
LGTM, only one comment, but it's a matter of taste.

jenkins_pipelines/environments/common/pipeline-build-validation-aws.groovy
NamelessOne91: To my understanding:

- We have a try/finally wrapping the whole pipeline

- Proxy stages are not wrapped in a try/catch at the stages level, a failure in one of them should make the pipeline skip ahead to the `finally` block.

 - Monitoring stages  are wrapped in a try/catch, so a fail here should keep the pipeline running to the end and mark it as red only once done and inside the `finally ` block.

NamelessOne91: Actually, I see your point too.

From sh docs 

> Normally, a script which exits with a nonzero status code will cause the step to fail with an exception

and I think this should be true also for calling `sh exit` with a code != 0
But I'm not 100% sure about it.

maximenoel8 (approved):
LGTM. Thank you !

nodeg (approved):

PR #1141, opened 2024-03-08
QE: Add containerized proxy to 5.0 BV

nodeg:
This will add the new containerized proxy to the 5.0 BV. I had to remove some of the options, since they are not available on the containerized proxy. See https://github.com/uyuni-project/sumaform/blob/90ef843cc7a62184328bfe23d13d4393c5957480/modules/proxy_containerized/main.tf

Depends on https://github.com/uyuni-project/sumaform/pull/1529

### DIFF
--- a/terracumber_config/tf_files/SUSEManager-5.0-build-validation-NUE.tf
+++ b/terracumber_config/tf_files/SUSEManager-5.0-build-validation-NUE.tf
@@ -185,7 +185,6 @@ module "server_containerized" {
   disable_firewall               = false
   allow_postgres_connections     = false
   skip_changelog_import          = false
-  create_first_user              = false
   mgr_sync_autologin             = false
   create_sample_channel          = false
   create_sample_activation_key   = false
--- a/terracumber_config/tf_files/SUSEManager-5.0-build-validation-NUE.tf
+++ b/terracumber_config/tf_files/SUSEManager-5.0-build-validation-NUE.tf
@@ -201,28 +201,24 @@ module "server_containerized" {
 
 }
 
-// WORKAROUND: We do the initial testing without a proxy and add it later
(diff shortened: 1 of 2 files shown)


Bischoff (approved):

terracumber_config/tf_files/SUSEManager-5.0-build-validation-NUE.tf
```diff
@@ -201,28 +200,24 @@ module "server_containerized" {
...
+  product_version    = "head"
+  name               = "pxy"
+  provider_settings = {
+    mac                = "aa:b2:92:42:00:a2"
+    memory             = 4096
+  }
+  server_configuration = {
+    hostname = "suma-bv-50-srv.mgr.suse.de"
+    username = "admin"
+    password = "admin"
```
szachovy, line 215: Not a fan of such a things in publicly available projects, can we at least encrypt it somehow and decrypt in the moment of authorization?

srbarrios: This is not a public repository.
Only SUSE employees have access to it.

srbarrios (approved):

terracumber_config/tf_files/SUSEManager-5.0-build-validation-NUE.tf
szachovy: Are you sure? I was able to access this comment from Incognito mode being signed out.

nodeg: > Not a fan of such a things in publicly available projects, can we at least encrypt it somehow and decrypt in the moment of authorization?
This was simply copy-pasted from the Uyuni/4.3 BV file.

PR #1149, opened 2024-03-13
Work around OOM problems

Bischoff:
Uyuni-master started having OOM kills on the proxy.

Working around the issue in order to get results, bu we should really understand if bug or normal evolution. If it's a bug, it should be fixed, and if on the contrary it's normal evolution, we should update our minimal requirements documentation.


srbarrios (approved):

nodeg (approved):
As a workaround, I am fine with the changes.

ktsamis (approved):

PR #1164, opened 2024-03-22
QE: Change the 5.0 BV confituration for testing

nodeg:
Temporary changes only for the 5.0 BV NUE configuration file.

- use the devel container until the 5.0 Beta 2 submissions are ready
- disable the mirror to verify a bug in the Product Wizard


nodeg:
Merging without approval since the changes are only temporary.

PR #1169, opened 2024-03-25
BV: Fixup for proxy `init_clients` command

nodeg:


nodeg:
Merging without approval since I tested this on the 5.0 BV controller and it is only a small fixup.

PR #1170, opened 2024-03-27
Test leap micro on PR testing for uyuni-product version

jordimassaguerpla:
Use Leap Micro for PR testing with containerized server and proxy.


srbarrios (approved):
It looks ok, but please rename the PR title to better describe that this applies only to uyuni-pr product version and so the PR test pipelines.

At some point later, it would be good to also change Uyuni Podman CI terraform file, to also use leapmicro55o image, we don't need to change the default for uyuni-master product version, but just for the value of image in the main.tf of that env.

cbosdo (commented):
Please set the k3s CI server and proxy images to `opensuse155o`: k3s fails to install on leap micro.

jordimassaguerpla:
> Please set the k3s CI server and proxy images to `opensuse155o`: k3s fails to install on leap micro.

I have only changed the uyuni-pr product in PR pipelines.

cbosdo (approved):

PR #1172, opened 2024-03-28
Add missing option necessary for pipeline

ktsamis:
Add missing option necessary for pipeline, we forgot to add it here, the SLE pipeline uses the BV.


ktsamis:
I didn't wait for review, it's a one liner that should fix the failing SLE testsuite

PR #1173, opened 2024-03-28
Change option in sle update pipeline

ktsamis:
Change option in sle update pipeline. Using proxy_traditional breaks the Rakefile, which needs adjusting. The Rakefile creates the feature files and run_sets from templates. We need to change pretty much every proxy feature file to proxy_traditional to make it work and until that refactor happens we need the pipeline to work.


ktsamis:
Merged without review to make the pipeline work and pass in the meantime

PR #1184, opened 2024-04-05
5.0 BV: Do not install the Salt bundle

nodeg:
This will disable installing the Salt bundle to all our SLE Micro Minions in the 5.0 BV.

See https://github.com/SUSE/spacewalk/issues/24074


nodeg:
Closed in favor of doing a Salt cleanup in the test suite.

PR #1185, opened 2024-04-05
BV-AWS: fix AMI images RegEx

NamelessOne91:
(Needs to be confirmed by running the pipeline with this change)

The previous regular expressions used to extract the images name from HTML might have been failing because of some incorrect escape characters.

This PR addresses the issue by introducing a slight refactor to avoid an HTTP call and using a different (hopefully more readable) RegEx.
Given that this pipeline definition is specifically for AWS BV, I think we can also make the RegEx more restrictive and have `EC2-Build` in it.


maximenoel8 (approved):
LGTM, thank you

PR #1188, opened 2024-04-08
HEAD CI proxy: Do not use `venv-salt-minion`

nodeg:
This will do not install the Salt bundle to our proxy VM on the HEAD CI, in order to be able to onboard the proxy as normal Minion again.

See https://github.com/SUSE/spacewalk/issues/24074


nodeg:
We do not want to get rid of the Salt bundle. Instead, we try to remove any sumaform leftovers via the test suite.

PR #1191, opened 2024-04-09
Add sles15sp5 to AWS BV 4.3

maximenoel8:
## What does this PR ? 
I created this PR after a discussion with Paolo. We don't cover sles15sp5 in AWS BV.

In this PR I added : 
 - add sles15sp5 client / minion / ssh minion
 - add migration client ( sles15sp5 )
 - Remove obsolete comment for BV repositories
 - Change order in main.tf 
 - Add domain to system using route53

### DIFF
--- a/terracumber_config/tf_files/SUSEManager-4.3-build-validation-AWS.tf
+++ b/terracumber_config/tf_files/SUSEManager-4.3-build-validation-AWS.tf
@@ -125,6 +125,10 @@ variable "NAME_PREFIX" {
   default = null
 }
 
+locals {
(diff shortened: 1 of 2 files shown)


NamelessOne91 (approved):
```
variable "NAME_PREFIX" {
  type = string
  default = null
}
```

having the default for this variable set to `null `makes Terraform fail because `null `cannot be used in strings interpolation. Maybe we can change the default to an empty string or use a ternary operator.

Other than that, LGTM. Thanks Maxime.

PR #1192, opened 2024-04-10
Fix proxy integration into our PR tests

srbarrios:
We were missing the deployment of the proxy container into our Uyuni PR test environment.


jordimassaguerpla (approved):
lgtm. thanks oscar

PR #1211, opened 2024-04-18
Adapt to new salt_obs_flavor nomenclature for salt_testenv

ycedres:
Change the `salt_obs_flavor` grain to the new expected values for the `salt_testenv` module.


meaksh (approved):
Looks good! BUT we need to wait until https://github.com/uyuni-project/sumaform/pull/1574 is merged.

PR #1223, opened 2024-04-22
Use SLES 15-SP4 for PR CI in Provo

cbosdo:
Using Leap 15.5 seems to break the CI as the proxy host couldn't be bootstrapped due to broken repo. Trying SLE 15 SP4 instead to unblock PR test.

### DIFF
--- a/terracumber_config/tf_files/tfvars/PR-testing-uyuni.tfvars
+++ b/terracumber_config/tf_files/tfvars/PR-testing-uyuni.tfvars
@@ -2,7 +2,7 @@
 
 IMAGE                  = "opensuse155-ci-pro"
 SERVER_IMAGE           = "leapmicro55o"
-PROXY_IMAGE            = "opensuse155o"
+PROXY_IMAGE            = "sles15sp4o"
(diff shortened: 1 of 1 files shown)


terracumber_config/tf_files/Uyuni-Master-PRV.tf
```diff
@@ -148,6 +148,7 @@ module "cucumber_testsuite" {
       }
       additional_packages = [ "venv-salt-minion" ]
       install_salt_bundle = true
+      image = "sles15sp4o"
```
srbarrios, line 151: Shouldn't Uyuni Proxy Non-containerized use Leap 15.5 as Base OS?

srbarrios (approved):

terracumber_config/tf_files/Uyuni-Master-PRV.tf
srbarrios: You infact wants to change this line https://github.com/SUSE/susemanager-ci/blob/master/terracumber_config/tf_files/tfvars/PR-testing-uyuni.tfvars#L5

cbosdo: Probably... but for now I have a weird issue with it and it's the containerized proxy that causes me troubles

PR #1226, opened 2024-04-23
Add salt migration client the BV list

maximenoel8:
Add salt migration client the BV list


NamelessOne91 (commented):
I missed this PR.

We do need to have it in the list, but I think the entry should use underscores. 
Aka: **salt_migration_minion**

PR #1231, opened 2024-05-10
BV: salt migration minion setup

NamelessOne91:
Takes care of not installing the salt bundle on the salt migration minion during deployment.

The minion should use only OS salt and auto-connect to the master.
During migration tests, we will install the salt bundle on it and switch it to use venv-salt, purging OS salt.

### DIFF
--- a/jenkins_pipelines/environments/common/pipeline-build-validation.groovy
+++ b/jenkins_pipelines/environments/common/pipeline-build-validation.groovy
@@ -209,15 +209,15 @@ def run(params) {
             /** Clients stages end **/
 
             /** Products and Salt migration stages begin **/
-            try {
-                stage('Products and Salt migration stages') {
-                    if(params.must_run_products_and_salt_migration_tests){
+            if(params.must_run_products_and_salt_migration_tests) {
+                try {
+                    stage('Products and Salt migration stages') {
                         clientMigrationStages()
-                    }                   
-                } 
-            } catch (Exception ex) {
-                println('ERROR: one or more migrations have failed')
-                products_and_salt_migration_stage_result_fail = true
+                    }
+                } catch (Exception ex) {
+                    println('ERROR: one or more migrations have failed')
+                    products_and_salt_migration_stage_result_fail = true
+                }
             }
             /** Products and Salt migration stages stages end **/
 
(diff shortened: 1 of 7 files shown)


terracumber_config/tf_files/Uyuni-Master-build-validation-NUE.tf
```diff
@@ -689,6 +689,7 @@ module "salt-migration-minion" {
   auto_connect_to_master  = true
   use_os_released_updates = false
   ssh_key_path            = "./salt/controller/id_rsa.pub"
+  install_salt_bundle = false
```
vandabarata, line 692: have you tried this locally? from what I remember, we added the salt bundle because there were issues when we didn't have it

NamelessOne91: I deployed the minion without salt-bundle in the 4.3.12 BV env, using terraform from the dedicated Jenkins workspace. It correctly contacts the server and its Salt key appears ready to be accepted.
Once that's done, it is onboarded and seems to have been working fine so far.

Maybe those are Uyuni specific issues ?

nodeg: FMPOV you can leave it in, Paolo, and test it in the Uyuni BV once this PR gets merged.

jenkins_pipelines/environments/common/pipeline-build-validation.groovy
```diff
@@ -209,15 +209,15 @@ def run(params) {
...
+                    stage('Products and Salt migration stages') {
                         clientMigrationStages()
-                    }                   
-                } 
-            } catch (Exception ex) {
-                println('ERROR: one or more migrations have failed')
-                products_and_salt_migration_stage_result_fail = true
+                    }
+                } catch (Exception ex) {
+                    println('ERROR: one or more migrations have failed')
```
nodeg, line 218: Above in `jenkins_pipelines/environments/common/pipeline-build-validation-aws.groovy` you use a different error message. Is this intentional?

terracumber_config/tf_files/Uyuni-Master-build-validation-NUE.tf
Bischoff: same fears as Vanda :)

NamelessOne91: Ok, this can stay parked here for a while and I can try this out in Uyuni BV to see what happens.
Tests will just do a venv-salt-minion installation anyway but nothing should change.

maximenoel8: `install_salt_bundle` option only install venv-salt-minion instead of salt-minion during the cloud init. This salt will then be used for the salt states and normally clear after the deployment.

PR #1245, opened 2024-05-21
Salt Shaker: Select proper package flavor when passing 'testsuite_dir' param

meaksh:
When running the Salt Shaker pipeline on SLMicro 6.0, we pass a `testsuite_dir` parameter to the pipeline in order to set `--directory` parameter to `salt-test` execution, in order to read the testsuite from a custom "rw" directory, but we were missing to propage the corresponding `--package-flavor` param in these cases, so the correct Python interpreter is used when running the testsuite.


vzhestkov (approved):
LGTM :+1:

PR #1248, opened 2024-05-23
BV: default false for highly disruptive stages

NamelessOne91:
Some of the tests which are enabled by setting these flags to **true** may leave the BV pipeline in a state which hinders following tests. Even more so when failing.

So, it  might be a be a good idea to have the default flag set to false to avoid running them by mistake.


Bischoff (approved):
Maybe change the title: ALL stages in BV are non-idempotent

vandabarata (approved):

maximenoel8 (approved):
LGTM

srbarrios (approved):
Thank you!

PR #1249, opened 2024-05-23
BV: remove containerization stages in 5.0 pipelines

NamelessOne91:
These stages execute some basic tests on a containerized proxy, that makes sense only for 4.3 BV pipelines.


Bischoff (approved):

maximenoel8 (approved):
lgtm

PR #1252, opened 2024-05-29
Define containers registry for the proxy

Bischoff:
To avoid having different registries in proxy and server, declare the required values on both.


mcalmer (approved):
Let's give it a try

Bischoff:
Ignoring sumaform validation

PR #1255, opened 2024-06-05
SL Micro 6.0: Deploy using install_salt_bundle = false, as we do for SLE Micro

srbarrios:
Otherwise, the bootstrap will fail.

### DIFF
--- a/terracumber_config/tf_files/SUSEManager-5.0-build-validation-NUE.tf
+++ b/terracumber_config/tf_files/SUSEManager-5.0-build-validation-NUE.tf
@@ -870,6 +870,7 @@ module "slmicro60-minion" {
 // WORKAROUND: Does not work in sumaform, yet
 //  additional_packages = [ "venv-salt-minion" ]
 //  install_salt_bundle = true
+  install_salt_bundle = false
 }
 
 module "sles12sp5-sshminion" {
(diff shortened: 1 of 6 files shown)


terracumber_config/tf_files/SUSEManager-5.0-build-validation-NUE.tf
```diff
@@ -870,6 +870,7 @@ module "slmicro60-minion" {
 // WORKAROUND: Does not work in sumaform, yet
 //  additional_packages = [ "venv-salt-minion" ]
 //  install_salt_bundle = true
+  install_salt_bundle = false
```
nodeg, line 873:
```suggestion
  install_salt_bundle = false
```

terracumber_config/tf_files/SUSEManager-5.0-build-validation-PRV.tf
```diff
@@ -1091,6 +1091,7 @@ module "slmicro60-minion" {
 // WORKAROUND: Does not work in sumaform, yet
 //  additional_packages = [ "venv-salt-minion" ]
 //  install_salt_bundle = true
+  install_salt_bundle = false
```
nodeg, line 1094:
```suggestion
  install_salt_bundle = false
```

nodeg (approved):
Some small suggestions. Otherwise LGTM.

terracumber_config/tf_files/SUSEManager-5.0-build-validation-NUE.tf
srbarrios: For now, I prefer to keep the whole block commented:
```
// WORKAROUND: Does not work in sumaform, yet
//  additional_packages = [ "venv-salt-minion" ]
//  install_salt_bundle = true
```

nodeg: OK, fine with me.

PR #1261, opened 2024-06-18
BV: Increase `repository_disk_size` to 3GB

nodeg:
We have more products to sync and therefore we need more disk spaces for the 4.3 and the Uyuni BV environment. For 5.0 the disk size is already at 3GB, see 
https://github.com/SUSE/susemanager-ci/blob/e710ef933aac05f29cf1e45fdbefd4abdb2905af/terracumber_config/tf_files/SUSEManager-5.0-build-validation-NUE.tf#L178


Bischoff (approved):
We only have 2.8 TB and 1.8 TB in Provo.

But it's okay to overcommit there...

srbarrios (approved):

PR #1264, opened 2024-06-26
Add liberty 9 to json generator and fix proxy repos

ktsamis:
Fixes https://github.com/SUSE/spacewalk/issues/24615

The things that this PR does:

- [x] Add Liberty 9 to json
- [x] Add sle micro 5.5 minion repos to the 5.0 proxy for after GA
- [x] Fixes the warning about missing IDs. This was the toughest to understand. Before we raised an exception and the script exited with 1. That caused it to NOT create the file. This can be a very usual case when we have both 4.3 and 5.0 IDs in the queue and it would exclude one version against the other depending on which one we specify we want. So there will always be IDs missing and the exit code would always/often be 1. We don't want that, we want a warning but I decided against introducing a new module (warnings module) and just print it.


NamelessOne91:
Added a commit with the unit test changes ;)

PR #1266, opened 2024-07-01
QE: use Rakefiles for BV migration stages

NamelessOne91:
Related to https://github.com/SUSE/spacewalk/issues/24233

Depends on https://github.com/uyuni-project/uyuni/pull/8989 and its  4.3 backport https://github.com/SUSE/spacewalk/pull/24679

Execute the migration stages using a Rake run set instead of the final feature file.
This allows to use the `cucumber:` namespace, which is responsible for creating and collect tests results.


maximenoel8 (approved):
LGTM

PR #1267, opened 2024-07-04
Salt Shaker: Add static IP address to all deployed VMs

meaksh:
Related to https://gitlab.suse.de/galaxy/infrastructure/-/merge_requests/911

PR #1269, opened 2024-07-05
Salt Shaker: Do not use avahi but static ip/hostname

meaksh:
We have enabled static ip addresses and hostname with:

https://gitlab.suse.de/galaxy/infrastructure/-/merge_requests/911
https://github.com/SUSE/susemanager-ci/pull/1267

But we were still missing to remove avahi from the TF files, as it is not needed now that we have static addresses / hostnames. Therefore this PR.


Bischoff (approved):
lgtm

PR #1275, opened 2024-07-19
BV: increase default timeout during client boostrap and disable the catch timeout method

maximenoel8:
### Increase the default timeout during client bootstrap
During my BV testing for 4.3 NUE, I had all my clients failing during bootstrap. This fail was related to the events taking up to 700 seconds. Currently the default timeout is set to 500 seconds for BV. I think this timeout is correct for everything except the client bootstrap ( the server can be slow at that time ).
That's why I added a temporary default timeout of about 800 for the client bootstrap stage ( I prefer to keep a ratio from default timeout if we need to change it later ).

### Disable catch timeout message mechanism
Also, the catch timeout method should not be needed anymore with the last server configuration. And I think this solution is bringing more problems than it solves. ( screenshot not available sometime when timing out ).
That's why I'm disabling this method for all the pipelines. Let's see if we see those timeout messages again. On a long term, we will need to remove completely this method from the testsuite.

### DIFF
--- a/jenkins_pipelines/environments/common/pipeline-build-validation.groovy
+++ b/jenkins_pipelines/environments/common/pipeline-build-validation.groovy
@@ -1,9 +1,9 @@
 def run(params) {
     timestamps {
         //Capybara configuration
-        def capybara_timeout= 60
+        def capybara_timeout = 60
         def default_timeout = 500
-        env.bootstrap_timeout        = 800
+        env.bootstrap_timeout = 800
 
         deployed = false
         env.resultdir = "${WORKSPACE}/results"
--- a/jenkins_pipelines/environments/common/pipeline-build-validation.groovy
+++ b/jenkins_pipelines/environments/common/pipeline-build-validation.groovy
@@ -484,7 +484,7 @@ def clientTestingStages() {
                     if (params.confirm_before_continue) {
                         input 'Press any key to start running the smoke tests'
                     }
-                    randomWait()
+//                    randomWait()
                     echo 'Run Smoke tests'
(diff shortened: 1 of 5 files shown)


jenkins_pipelines/environments/common/pipeline-build-validation.groovy
```diff
@@ -472,7 +472,9 @@ def clientTestingStages() {
                     }
                     randomWait()
                     echo 'Bootstrap clients'
-                    res_init_clients = sh(script: "./terracumber-cli ${common_params} --logfile ${resultdirbuild}/testsuite.log --runstep cucumber --cucumber-cmd 'unset ${temporaryList.join(' ')}; ${env.exports} cd /root/spacewalk/testsuite; rake cucumber:build_validation_init_client_${node}'", returnStatus: true)
+                    // Increase by 1.6 the default timeout (800 seconds) to manage the events taking up to 750 seconds during bootstrap
+                    def temporary_timeout = env.default_timeout * 1.6
```
srbarrios, line 476: What do you think if we define a `env.bootstrap_client_timeout = 800` just below `env.default_timeout = 500` and use it here?

Otherwise this calculation becomes a [magic number](https://refactoring.guru/es/replace-magic-number-with-symbolic-constant).

maximenoel8: Yes it's probably better. If I remember correctly, we had this default timeout so high for bootstrap reason. We may even be able to reduce a bit the max timeout if we separate it from the bootstrap features. But let's focus on getting BV stable first

PR #1285, opened 2024-07-29
Change registry to one that does not need auth

ktsamis:
Change registry to one that does not need auth


ktsamis:
we won't change the registry but add authentication

PR #1287, opened 2024-07-31
Don't use the default venv-salt-minion for salt migration minon

maximenoel8:
## What does this PR do ?

In 4.3 BV, salt-migration is always failing to bootstrap. I was able to fix the issue by rebooting venv-salt-minion service but I was surprise we were using venv-salt-minion and not OS salt.
From my understanding, we want to bootstrap this minion using salt and then do the migration ? 

I need some feedback from @NamelessOne91.
If we want to bootstrap using salt, those changes will fix the deployment.
If we don't specify the `install_salt_bundle`, it will be set to `true` by default.
With `auto_connect_to_master` set to true and `install_salt_bundle` set to true, we will bootstrap the minion using venv-salt-minion  :
https://github.com/uyuni-project/sumaform/blob/676c3aaacd724ef454f6da2bcad9f3129e76591d/salt/minion/init.sls#L50-L81


NamelessOne91 (approved):
Thanks Maxime

PR #1290, opened 2024-08-08
Add proxy retail test to qe update to automatically

maximenoel8:
## What does this PR change?

In the current SLE update pipeline, the retail proxy feature is not being executed, which means that bind and DHCP packages are not automatically tested. To eliminate the need for manual testing of a feature that is already validated in Build Validation, this PR introduces the option to run the retail proxy feature. Since we do not have a terminal or build clients available, the scope is limited to running `features/build_validation/retail/proxy_traditional_branch_network.feature`. The option will be disabled by default and should be enabled only when relevant packages are involved.

## Ideas for Further Improvement

Deploying sle15sp4_terminal and sle15sp4_buildhost as additional clients could enhance testing for TFTP packages, potentially saving time. However, this would require changes to the infrastructure. @Bischoff, is there enough space on PRV to accommodate these two additional clients?

### DIFF
--- a/jenkins_pipelines/environments/manager-4.3-qe-sle-update
+++ b/jenkins_pipelines/environments/manager-4.3-qe-sle-update
@@ -35,6 +35,7 @@ node('sumaform-cucumber-provo') {
             booleanParam(name: 'must_create_bootstrap_repos', defaultValue: true, description: 'Create bootstrap repositories'),
             booleanParam(name: 'must_boot_node', defaultValue: true, description: 'Bootstrap Node'),
             booleanParam(name: 'must_run_tests', defaultValue: true, description: 'Run Smoke Tests'),
+            booleanParam(name: 'must_prepare_retail', defaultValue: false, description: 'Prepare and run Retail'),
             booleanParam(name: 'must_run_containerization_tests', defaultValue: false, description: 'Run Containerization Tests'),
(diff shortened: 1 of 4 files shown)


srbarrios (changes requested):
I like the idea of covering a sub-set of Retail tests.

⚠️ Keep in mind that DNS/DHCP is installed and configured by salt formulas as part of the Proxy/RBS Host in 4.3 (at least if we don't use the containerized flavor of the RBS 4.3), while in 5.0, DNS/DHCP is installed and configured by terraform in a different external host, meaning it's out of the responsibilities of the product.
So, regarding this PR, we miss this part on terracumber_config/tf_files/SUSEManager-5.0-SLE-update.tf

```
module "dhcp-dns" {
  source             = "./modules/dhcp_dns"
  base_configuration = module.base_core.configuration
  name               = "dhcp-dns"
  image              = "opensuse155o"
  private_hosts = [
    module.proxy_containerized.configuration
  ]
  hypervisor = {
    host        = "riverworld.mgr.prv.suse.net"
    user        = "root"
    private_key = file("~/.ssh/id_rsa")
  }
}
```

maximenoel8:
I removed 5.0 for the retail scope.

Bischoff:
> I removed 5.0 for the retail scope.

Hmmm, if you enable the DHCP and DNS server, you could make it work on 5.0 as well.

PR #1291, opened 2024-08-09
Add possibility to use file to declare the sle update reference number

maximenoel8:
## What does this PR ? 

Add the option to specify mi IDs in a file instead of providing them directly on the command line.

### File example 
File name : mis_12082024
Content : 
```
35198
32007
35176
34880
35142
35177
35114
```
Command : `maintenance_json_generator.py -f mis_12082024`

### DIFF
--- a/jenkins_pipelines/scripts/json_generator/maintenance_json_generator.py
+++ b/jenkins_pipelines/scripts/json_generator/maintenance_json_generator.py
@@ -212,21 +212,22 @@ def parse_cli_args() -> argparse.Namespace:
     parser.add_argument("-f", "--file", required=False, dest="file", help="Path to a file containing MI IDs separated by newline character", action='store')
     parser.add_argument("-e", "--no_embargo", dest="embargo_check", help="Reject MIs under embargo",  action='store_true')
     args: argparse.Namespace = parser.parse_args()
-
-    # If file argument is provided, read MI IDs from the file
-    if args.file:
-        try:
-            with open(args.file, 'r') as file:
-                file_content = file.read().strip()
-                file_mi_ids = file_content.split()
-                if args.mi_ids:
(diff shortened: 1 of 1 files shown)


jenkins_pipelines/scripts/json_generator/maintenance_json_generator.py
```diff
@@ -205,9 +209,23 @@ def parse_cli_args() -> argparse.Namespace:
...
     parser.add_argument("-i", "--mi_ids", required=False, dest="mi_ids", help="Space separated list of MI IDs", nargs='*', action='store')
+    parser.add_argument("-f", "--file", required=False, dest="file", help="Path to a file containing MI IDs separated by newline character", action='store')
     parser.add_argument("-e", "--no_embargo", dest="embargo_check", help="Reject MIs under embargo",  action='store_true')
-
     args: argparse.Namespace = parser.parse_args()
+
+    # If file argument is provided, read MI IDs from the file
+    if args.file:
+        try:
+            with open(args.file, 'r') as file:
```
NamelessOne91, line 219: IMO, it would be better to move this file-parsing logic to a separate function that takes a file path and returns the MI IDs. 
We can check if args.file is set later (e.g in the main block) and call it there.

That's mostly to make it easier to test it and let  the **parse_cli_args** function be responsible of only one thing.
I can take care of adding a few unit tests for it.

NamelessOne91 (changes requested):
Overall, I'm fine with addiction of the new flag - it is indeed a useful thing to have - but I'd like to have the file-parsing code added in a way that makes it easy to write unit tests for it and preserve the ones for parse_cli_args.

If we intend to use proper logging, there are a few print calls around which would make sense to replace as well - see line 255 as an example.

PR #1298, opened 2024-08-14
Update non mu list

maximenoel8:
Remove clm declaration for the non MU repositories in BV. We are now doing it in CI


NamelessOne91 (commented):
Thanks Maxime

PR #1299, opened 2024-08-15
Add unit tests for file flag functions

NamelessOne91:
This PR

- replaces print calls with proper logging
- lets the program fail when there's an error reading a file provided through **-f** or **--file** flag
- adds unit tests for the functions responsible of reading such file and merging MI IDs with the ones provided with the **-i** flag


maximenoel8 (approved):
LGTM. Thank you !

jenkins_pipelines/scripts/json_generator/ibs_osc_client.py
```diff
@@ -68,7 +69,7 @@ def _parse_embargo_date(self, text: str) -> date:
             try:
                 return datetime.strptime(text, format).date()
             except:
-                print(f"\tInvalid date format {format} for OBS:EmbargoDate value {text}")
+                logging.error(f"\tInvalid date format {format} for OBS:EmbargoDate value {text}")
         
         raise ValueError(f"Failed to parse OBS:EmbargoDate value {text} with all available date formats")
```
szachovy, line 74: You can just raise `ValueError` with that message instead, optional

jenkins_pipelines/scripts/json_generator/maintenance_json_generator.py
```diff
@@ -215,12 +215,8 @@ def parse_cli_args() -> argparse.Namespace:
...
 
 def read_mi_ids_from_file(file_path: str) -> list[str]:
-    try:
-        with open(file_path, 'r') as file:
-            return file.read().strip().split()
-    except Exception as e:
-        logging.error(f"Error reading file: {e}")
-        return []
+    with open(file_path, 'r') as file:
+        return file.read().strip().split()
```
szachovy, line 219: Why exception is no longer needed? What would happen if we get let say permission denied?

jenkins_pipelines/scripts/tests/test_maintenance_json_generator.py
```diff
@@ -138,6 +142,33 @@ def test_validate_and_store_results(self, mock_print):
...
 
+    def test_read_mi_ids_from_file(self):
+        file_ids: list[str] = read_mi_ids_from_file(_TEST_MI_IDS_FILE_PATH)
+        self.assertEqual(file_ids, ['11111', '22222', '33333'])
+
+        self.assertRaises(OSError, read_mi_ids_from_file, 'file_not_found.txt')
+
+    def test_merge_mi_ids(self):
+        # no ids at all
+        sys.argv= ['maintenance_json_generator.py', '-v', '50']
```
szachovy, line 153: everywhere space after `sys.argv`

jenkins_pipelines/scripts/tests/test_maintenance_json_generator.py
```diff
@@ -138,6 +142,33 @@ def test_validate_and_store_results(self, mock_print):
...
+        # only mi ids file
+        sys.argv= ['maintenance_json_generator.py', '-v', '50', '-f', _TEST_MI_IDS_FILE_PATH, '-e']
+        args = parse_cli_args()
+        ids: list[str] = merge_mi_ids(args)
+        self.assertEqual(ids, ['11111', '22222', '33333'])
+        # ids both from flag and file
+        sys.argv= ['maintenance_json_generator.py', '-v', '50', '-f', _TEST_MI_IDS_FILE_PATH, '-i', '1234', '5678', '-e']
+        args = parse_cli_args()
+        ids: list[str] = merge_mi_ids(args)
+        self.assertEqual(ids, ['1234', '5678', '11111', '22222', '33333'])
```
szachovy, line 171: all `-i` values are smaller than this in file, I would suggest mixing them.

jenkins_pipelines/scripts/tests/test_maintenance_json_generator.py
```diff
@@ -138,6 +142,33 @@ def test_validate_and_store_results(self, mock_print):
         # cleanup
         remove(test_output_file)
 
+    def test_read_mi_ids_from_file(self):
+        file_ids: list[str] = read_mi_ids_from_file(_TEST_MI_IDS_FILE_PATH)
```
szachovy, line 146: Why exacly `_TEST_MI_IDS_FILE_PATH` is a global variable? Can we do it statically as it is with `test_output_file`?

szachovy (commented):
Optional, feel free to merge

jenkins_pipelines/scripts/tests/test_maintenance_json_generator.py
szachovy: Especially if you already created this file, any other parameter would fail tests.

ktsamis (commented):
Looks good to me but I see Wiktor's questions and I would approve after we have all clarified

jenkins_pipelines/scripts/json_generator/maintenance_json_generator.py
NamelessOne91: The previous version captured all possible exceptions, logged an error and returned an empty list.
This meant that any error occurring with the file would not stop the script execution.

IMO is better to just let any exception regarding the file (permission issues, file not found...) propagate up the stack and have the script crash in such cases, so it is evident there's an issue and we do not risk using the results as if everything worked fine when it actually did not.

jenkins_pipelines/scripts/json_generator/ibs_osc_client.py
NamelessOne91: The logic of this function is to try and parse the text we get in the date field with all the available formats we know to be used.
If one works, cool, we return the parsed date. If it does not, I want to have a log of the failure but not raise an exception  - otherwise the function will stop at that point and other formats won't be tried.

If none of them works, then line 73 will raise the ValueError and lead to a crash. At this point the previous logs would help knowing what the text was and what date formats we tried.

jenkins_pipelines/scripts/json_generator/maintenance_json_generator.py
szachovy: Theoretically right, every exception is already handled in built-in. I just wanted to be reassured that this step would not pass in any border case like corrupted records, file but exception not thrown. But it should have been already done in the previous solution.

Resolving

jenkins_pipelines/scripts/json_generator/ibs_osc_client.py
szachovy: I missed it is `for` loop, thank you for clarification.

Resolving

jenkins_pipelines/scripts/tests/test_maintenance_json_generator.py
NamelessOne91: If that's because you see ordered results and you want to test whether that order is kept with different IDs, I can reassure that there's no ordering involved in these steps .


All it does is put the IDS coming from  the -i flag before the ones coming from the file. The actual numerical order does not matter here.
The keys will be ordered once the final JSON output is produced.

szachovy: ok, but where we are testing that it sorts correctly?

szachovy: Anyway, we have a new feature we should test it. I am sure that parameter of this function works correctly but given that we are parsing IDs from a two parameters, or even more combinations can be involved that I am not aware of. I would check in the end if IDs from file `<1, 4, 5>` and IDs from `-i` parameter like `-i 2, 6 -i 3` (yes, double `i` parameter) it returns `1, 2, 3, 4, 5, 6`.

I would do that, but if you consider it unnecessary, you can close it.

PR #1300, opened 2024-08-15
Unset disable clients before sanity check

maximenoel8:
## Context
When nodes were disable in the minion list, they were still check by the sanity check

## What does this PR
Correctly unset the disable nodes during sanity check stage.

### DIFF
--- a/jenkins_pipelines/environments/common/pipeline-build-validation.groovy
+++ b/jenkins_pipelines/environments/common/pipeline-build-validation.groovy
@@ -75,7 +75,7 @@ def run(params) {
 
             stage('Sanity check') {
                 def nodesHandler = getNodesHandler()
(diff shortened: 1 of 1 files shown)


jenkins_pipelines/environments/common/pipeline-build-validation.groovy
```diff
@@ -542,7 +543,7 @@ def getNodesHandler() {
         BootstrapRepositoryStatus[node] = 'NOT_CREATED'
         CustomChannelStatus[node]       = 'NOT_CREATED'
     }
-    return [nodeList:nodeListWithDisabledNodes, envVariableList:envVar, envVariableListToDisable:envVarDisabledNodes, CustomChannelStatus:CustomChannelStatus, BootstrapRepositoryStatus:BootstrapRepositoryStatus]
+    return [nodeList:nodeListWithDisabledNodes, envVariableList:envVar, envVariableListToDisable:envVarDisabledNodes.join(' '), CustomChannelStatus:CustomChannelStatus, BootstrapRepositoryStatus:BootstrapRepositoryStatus]
```
nodeg, line 546: So with this, we get an "empty" list of nodes if we do not want to disable any, and in the terracumber call above, we then unset the environment variable. I like it.

nodeg (approved):

PR #1304, opened 2024-08-16
Fixes for the Uyuni BV

nodeg:
This PR

- ~fixes the main disk image size~
- removes the CentOS 7 client
- explicitly defines the server OS image version
- removes `catch_timeout_message = false` from the controller

for the Uyuni BV.

### DIFF
--- a/terracumber_config/tf_files/Uyuni-Master-build-validation-NUE.tf
+++ b/terracumber_config/tf_files/Uyuni-Master-build-validation-NUE.tf
@@ -168,6 +168,7 @@ module "server_containerized" {
   base_configuration = module.base_core.configuration
   product_version    = "uyuni-master"
   name               = "srv"
+  image              = "leapmicro55o"
   provider_settings = {
     mac                = "aa:b2:93:02:01:a1"
     memory             = 40960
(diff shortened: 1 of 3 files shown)


vandabarata (approved):

terracumber_config/tf_files/Uyuni-Master-build-validation-NUE.tf
```diff
@@ -176,9 +177,7 @@ module "server_containerized" {
   }
 
   server_mounted_mirror = "minima-mirror-ci-bv.mgr.suse.de"
-  main_disk_size        = 20
-  repository_disk_size  = 3072
-  database_disk_size    = 150
+  main_disk_size = 3000
```
Bischoff, line 180: this is incrorrect, for 2 reasons:

* using separate partitions for packages and database is good practice
* the problem in not on main disk (/) but on /var, so it won't work

the problem seems to arise from the initial partitioning of a sle micro

Bischoff (changes requested):
nope :smile_cat:

terracumber_config/tf_files/Uyuni-Master-build-validation-NUE.tf
nodeg: Makes sense. Thank you for the comment. We discussed this in a huddle and concluded that I will revert this change for now until we know how to fix it.

Bischoff (approved):

PR #1313, opened 2024-08-20
Create specific non MU channel task file for 4.3 and for 5.0

maximenoel8:
## Context 
The new Appstream support for 5.0 make the clm use during non MU stages useless.
Those features have been removed from 5.0 testsuite.

For 4.3, we don't have the new appstream support so we still need to create those CLM projects.

## Solution

 - Create two separate non MU channels task for 4.3 and 5.0
 - Point to the correct non MU channels file at the parameter pipeline description level


Bischoff (approved):
thank you

srbarrios (approved):
The fix lgtm.

In a previous refactor, it seems we started to use json format to map the cucumber features to trigger depending the client.

I missed it sorry, personally I think it would be worth to consider an aligment in the file format we use for that purpose, as currently we have YAML in some parts of the pipeline, but JSON in others, making harder to maintain and understand.

maximenoel8:
If I remember correctly ( it was more then 2 years ago ), we choose json format for the groovy pipeline because it is natively supported where yaml need plugin. If I'm correct I was never able to make it work.

NamelessOne91 (approved):
Thanks

PR #1323, opened 2024-08-26
Fix registry url

ktsamis:
Fix registry url, the rest gets added in the code of the mgr tools.


nodeg (approved):

nodeg:
See what Michele said about the URL:

>>For uyuni it was always uyuni/server[something] which caused the duplicate "uyuni" in the path

>TLDR; yes.
>Default image name is suse/manager/5.0/x86_64/server[something] (for mgradm from IBS), so if you don't want that parte append, you can specify the image name (from CLI or config file) (edited) 


See https://suse.slack.com/archives/C02CLB6AGJ3/p1724666740467119?thread_ts=1724567586.158589&cid=C02CLB6AGJ3

vandabarata (approved):
thank you so much :green_heart:

PR #1324, opened 2024-08-26
Use latest tag in SLE testsuite containers

ktsamis:
Use latest tag in SLE testsuite containers because otherwise it's trying 5.0.0 and the MI registry of course doesn't have this.


ktsamis:
merging this to try it out

PR #1326, opened 2024-08-28
fix missing nmcli in proxy

jordimassaguerpla:
nmcli is being used for testing, see https://github.com/uyuni-project/uyuni/blob/ce8e2e104d4bf181e949e0a57275e2657a9421b8/testsuite/features/build_validation/retail/proxy_container_branch_network.feature

nmcli belongs to the NetworkManager RPM.

However, the proxy vm does not have that package installed.

This PR installs this package, so we can use it during the tests.


Bischoff (changes requested):
No.

We use nmcli because the base image for containerized proxy is a a micro os.

PR #1327, opened 2024-08-28
Fix proxy in PR testing

jordimassaguerpla:
We need to use the micro version of openSUSE Leap. This is what it is expected and what we have already in the "podman pipeline".

Otherwise, we would need to install NetworkManager, which is not in the openSUSE Leap "non-micro" version.

### DIFF
--- a/terracumber_config/tf_files/tfvars/PR-testing-uyuni.tfvars
+++ b/terracumber_config/tf_files/tfvars/PR-testing-uyuni.tfvars
@@ -2,7 +2,7 @@
 
 IMAGE                  = "opensuse155-ci-pro"
 SERVER_IMAGE           = "leapmicro55o"
-PROXY_IMAGE            = "opensuse155o"
+PROXY_IMAGE            = "leapmicro55o"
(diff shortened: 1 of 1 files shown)


Bischoff (approved):
thanks

srbarrios (approved):

terracumber_config/tf_files/tfvars/PR-testing-uyuni.tfvars
```diff
@@ -2,7 +2,7 @@
 
 IMAGE                  = "opensuse155-ci-pro"
 SERVER_IMAGE           = "leapmicro55o"
-PROXY_IMAGE            = "opensuse155o"
+PROXY_IMAGE            = "leapmicro55o"
 IMAGES                 = ["rocky9o", "opensuse155o", "opensuse155-ci-pro", "ubuntu2204o", "sles15sp4o", "leapmicro55o"]
```
srbarrios, line 6: Do we use "opensuse155o" somewhere?
Otherwise we will be downloading it for nothing

jordimassaguerpla: good point. Let me create a separate PR

PR #1329, opened 2024-08-29
Fix container URL in the HEAD CI

nodeg:
Currently we see duplicate parts of the URL due to some parts will get auto-appended with the `mgr*` tools.

```bash
ing to pull registry.suse.de/devel/galaxy/manager/head/containerfile/suse/manager/5.0/x86_64/suse/manager/5.0/x86_64/server:latest...
or: initializing source docker://registry.suse.de/devel/galaxy/manager/head/containerfile/suse/manager/5.0/x86_64/suse/manager/5.0/x86_64/server:latest: reading manifest latest in registry.>
or: exit status 125
```
The URL contains 2 times `/suse/manager/5.0/x86_64/`.

### DIFF
--- a/terracumber_config/tf_files/SUSEManager-Head-NUE.tf
+++ b/terracumber_config/tf_files/SUSEManager-Head-NUE.tf
@@ -143,7 +143,7 @@ module "cucumber_testsuite" {
       main_disk_size = 500
       login_timeout = 28800
       runtime = "podman"
-      container_repository = "registry.suse.de/devel/galaxy/manager/head/containerfile/suse/manager/5.0/x86_64"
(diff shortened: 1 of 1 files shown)


ktsamis (changes requested):
I don't think the proxy needs changing, I had the same issue here https://github.com/SUSE/susemanager-ci/pull/1325

Bischoff (approved):
only in this file?

nodeg:
Interesting. I will ask Michele about that. I think it should work the same with `mgrpxy` but this could be due to the current bug.

nodeg:
>only in this file?

Yes. I did not want to couch the 5.0 BV file, since we need to change it anyway with the next BV. Uyuni Podman is already correct.

ktsamis (approved):

PR #1330, opened 2024-08-29
Enable Liberty 9 again in our BV

nodeg:
This enables the Liberty 9 Minions again, but without installing the Salt bundle beforehand. The SUMA server will take care of this. We cannot install it anymore since we do not register Liberty 9 with RMT anymore.

See https://github.com/uyuni-project/sumaform/pull/1663


vandabarata (approved):

nodeg:
I enabled the commented out Salt bundle again since we have the client tools repo available.

PR #1331, opened 2024-08-30
[salt shaker] Fix Debian 11 pipeline fails

m-czernek:
In this PR, I:

- Change the Debian 11 default parameter to not use the `use_previous_terraform_state` value. This is because on Debian 11, this causes fails, most likely due to problems with TF environment reprovisioning. The main difference seems to be that when `use_previous_terraform_state` is used, the environment resources do not get the `taint: true` property (even when explicitly specified in the job), which means some of the KVM resources might not get recreated.
- Removing Debian 10  and SLES15 SP1 Jenkins jobs, which are now deprecated and not tested. Those jobs now exist in the https://ci.suse.de/view/zz-DELETEME/ view, until they get deleted.


vzhestkov (approved):
LGTM :+1:

PR #1337, opened 2024-09-04
Use shorter registry identifiers

Bischoff:
Do not specify completely the image name, let mgradm and mgrpxy choose


srbarrios (approved):
LGTM.
Although this refactor I'm not sure if it's already in place on Manager-5.0, or that only applies to main for now.

Bischoff:
> LGTM. Although this refactor I'm not sure if it's already in place on Manager-5.0, or that only applies to main for now.

5.0 CI is not operational yet, no, so it's HEAD only for the time being.

PR #1361, opened 2024-09-13
Add product_version to controller in BV and QE Maintenance

maximenoel8:
Add the new product version variable.

This variable select the correct ruby version


srbarrios:
In order to fix the current issue, LGTM.
But, if someone can spend some time on it...
I think we need to move product_version to the base module, and so we use that value in all the modules that require it, through module.base.configuration

srbarrios (approved):

PR #1362, opened 2024-09-13
Replace sets with lists as values for tools dictionaries

NamelessOne91:
This should solve https://github.com/SUSE/spacewalk/issues/25032 

Python sets guarantee the absence of duplicate values but the actual order of items at runtime is not guaranteed to stay the same. 
For lists the reverse is true.

Given the same inputs, we want each run to produce exactly the same JSON output so we'll to stick with lists and make sure through other means that we do not insert duplicate values in the dictionaries.

### DIFF
--- a/jenkins_pipelines/scripts/json_generator/maintenance_json_generator.py
+++ b/jenkins_pipelines/scripts/json_generator/maintenance_json_generator.py
@@ -192,9 +192,12 @@ v50_nodes: dict[str, set[str]] = {
 }
 v50_nodes.update(merged_client_tools)
 
-nodes_by_version: dict[str, dict[str, set[str]]] = {
-    "43": v43_nodes,
-    "50": v50_nodes
+v43_nodes_sorted: dict[str, list[str]] = {k:sorted(v) for k,v in v43_nodes.items()}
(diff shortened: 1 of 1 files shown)


maximenoel8 (approved):
LGTM, could you create an other PR / issue to cover the duplication risk please ?

NamelessOne91:
> LGTM, could you create an other PR / issue to cover the duplication risk please ?

We can add extra code to filter out duplicates, but I'd avoid that for the time being.
Lists is  actually what we had in the beginning and it is mostly a matter of double checking when inserting new values in the static dictionaries.
Also, I think in the end duplicate values will just overwrite their own key in the final JSON (will have to check, though).

srbarrios:
@NamelessOne91 Did you try to keep using _set_, but just return `sorted(<your_set>)`, so it returns a list that will have always the same order?

NamelessOne91:
> @NamelessOne91 Did you try to keep using _set_, but just return `sorted(<your_set>)`, so it returns a list that will have always the same order?

Good point !
I'll adapt the code to use this.

srbarrios (approved):

PR #1370, opened 2024-09-19
Remove KVM host from HEAD and Master

cbosdo:
Not removing it from the test projects since those could test 4.3 still. Each squad should decide what to do with it.


cbosdo:
Related code [removal PR](https://github.com/uyuni-project/uyuni/pull/9276) being worked on, but the tests can already be disabled in the CIs.

cbosdo:
The test failure looks fishy to me... is that expected or did I remove a bit too much?

srbarrios (approved):

srbarrios:
> The test failure looks fishy to me... is that expected or did I remove a bit too much?

It looks unrelated to me. 👀

PR #1391, opened 2024-09-26
Add image declarations to different bases for 5.0 BV PRV

maximenoel8:
## Context

5.0 BV PRV is currently because of this error: 
```
Error: Can't retrieve base volume with name 'suma-bv-50-libertylinux9o': virError(Code=50, Domain=18, Message='Storage volume not found: no storage vol with matching name 'suma-bv-50-libertylinux9o'')
```

This error is caused by not declaring the image at the base level and then using it at the module level.

## What does this PR ?

Correctly declare this images and base for all minion in 5.0 BV PRV.


Bischoff (approved):
thank you

NamelessOne91 (approved):

PR #1395, opened 2024-09-30
Filter the output of the terraform apply command to make it more readable

srbarrios:
This PR change (**ONLY in pipeline.groovy for now**) the output of the terraform apply command to make it more readable.

Example:

Previous output:
```
[2024-09-30T08:07:09.425Z] Resource instance example_resource has been created.
[2024-09-30T08:18:36.562Z] module.cucumber_testsuite.module.controller.module.controller.module.host.null_resource.provisioning[0] (remote-exec): <some_random_commands>
```

New output:
```
[08:07:09] Resource instance example_resource has been created.
[08:18:36] cucumber_testsuite.controller (remote-exec): <some_random_commands>
```

### DIFF
--- a/jenkins_pipelines/environments/common/pipeline.groovy
+++ b/jenkins_pipelines/environments/common/pipeline.groovy
@@ -94,7 +94,7 @@ def run(params) {
                             break;
                     }
                 }
-                sh "set +x; source /home/jenkins/.credentials set -x; export TF_VAR_CUCUMBER_GITREPO=${params.cucumber_gitrepo}; export TF_VAR_CUCUMBER_BRANCH=${params.cucumber_ref}; export TERRAFORM=${params.terraform_bin}; export TERRAFORM_PLUGINS=${params.terraform_bin_plugins}; ./terracumber-cli ${common_params} --logfile ${resultdirbuild}/sumaform.log ${env.TERRAFORM_INIT} ${env.TERRAFORM_TAINT} --sumaform-backend ${params.sumaform_backend} --runstep provision"
+                sh "set +x; source /home/jenkins/.credentials set -x; export TF_VAR_CUCUMBER_GITREPO=${params.cucumber_gitrepo}; export TF_VAR_CUCUMBER_BRANCH=${params.cucumber_ref}; export TERRAFORM=${params.terraform_bin}; export TERRAFORM_PLUGINS=${params.terraform_bin_plugins}; ./terracumber-cli ${common_params} --logfile ${resultdirbuild}/sumaform.log ${env.TERRAFORM_INIT} ${env.TERRAFORM_TAINT} --sumaform-backend ${params.sumaform_backend} --runstep provision | sed -E 's/([^.]+)module\\.([^.]+)\\.module\\.([^.]+)(\\.module\\.[^.]+)?(\\[[0-9]+\\])?(\\.module\\.[^.]+)?(\\.[^.]+)?(.*)/\\1\\2.\\3\\8/'"
                 deployed = true
                 // Collect and tag Flaky tests from the GitHub Board
                 def statusCode = sh script:"./terracumber-cli ${common_params} --logfile ${resultdirbuild}/testsuite.log --runstep cucumber --cucumber-cmd 'cd /root/spacewalk/testsuite; ${env.exports} rake utils:collect_and_tag_flaky_tests'", returnStatus:true


Bischoff (changes requested):
```
 while read -r line; do echo \"$line\" | 
 ```
 I suppose you could simply remove that step in the UNIX pipe ?

File: jenkins_pipelines/environments/common/pipeline.groovy [hunk 1]
```diff
@@ -94,7 +94,7 @@ def run(params) {
...
+                sh "set +x; source /home/jenkins/.credentials set -x; export TF_VAR_CUCUMBER_GITREPO=${params.cucumber_gitrepo}; export TF_VAR_CUCUMBER_BRANCH=${params.cucumber_ref}; export TERRAFORM=${params.terraform_bin}; export TERRAFORM_PLUGINS=${params.terraform_bin_plugins}; ./terracumber-cli ${common_params} --logfile ${resultdirbuild}/sumaform.log ${env.TERRAFORM_INIT} ${env.TERRAFORM_TAINT} --sumaform-backend ${params.sumaform_backend} --runstep provision | while read -r line; do echo \"$line\" | sed -E 's/\\[([0-9]{4}-[0-9]{2}-[0-9]{2})T([0-9]{2}:[0-9]{2}:[0-9]{2})\\.[0-9]+Z\\] module\\.([^.]+)\\.module\\.([^.]+)\\.module\\.([^.]+)\\.module\\.[^.]+\\.null_resource\\.[^.]+\\[[0-9]+\\] (.*)/[\\2] \\3.\\4 \\6/'; done"
```
szachovy, line 97: Good idea to get rid of all this `module.` that nobody follows, however I would stay with the original timestamp.

Bischoff: Agree

srbarrios:
> ```
> while read -r line; do echo \"$line\" | 
> ```
> 
> I suppose you could simply remove that step in the UNIX pipe ?

This while reads each line of the output from terraform apply one by one, it parse it with sed and prints the resulted line.
Also, ff we remove this part, it will process it all the lines together, and that will imply no output in the console until the command finish.

jenkins_pipelines/environments/common/pipeline.groovy
srbarrios: From [2024-09-30T08:07:09.425Z] to [08:07:09].

- The date is part of all our Jenkins jobs, and this replacement it will only happen in pipelines, so what's the value of adding it for every log line?
- The milliseconds add very little value, if we take into account that the log lines are printed chronologically.

Do you really still prefer to have it complete?

Bischoff:
> > ```
> > while read -r line; do echo \"$line\" | 
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > I suppose you could simply remove that step in the UNIX pipe ?
> 
> This `while` statement reads each line of the output from `terraform apply` one by one, it parse it with sed and prints the resulted line. Also, ff we remove this part, it will process it all the lines together, and that will imply no output in the console until the command finish.

sed can process multiple lines...

To me this looks completely useless.

srbarrios:
> > > ```
> > > while read -r line; do echo \"$line\" | 
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > I suppose you could simply remove that step in the UNIX pipe ?
> > 
> > 
> > This `while` statement reads each line of the output from `terraform apply` one by one, it parse it with sed and prints the resulted line. Also, ff we remove this part, it will process it all the lines together, and that will imply no output in the console until the command finish.
> 
> sed can process multiple lines...
> 
> To me this looks completely useless.

I'm gonna give a try without it.

jenkins_pipelines/environments/common/pipeline.groovy
szachovy: Milliseconds maybe not, but `year-month-day` are useful when someone needs a help with debugging on let say master branch from around this `year-month-day-hour`, if you run code next day in the similar hour you might get the same log but on the different codebase if master gets updated on the new commits, it may rise confusions in my opinion, and I do not support removing it.

szachovy: And I won't rely on the jenkins info on the screenshot because this run may disappear any time, but plain text from error log not, attaching this kind of screenshot to bugzilla is not an attractive option indeed.

srbarrios: Ok, I will keep year-month-day and filter milliseconds 👍🏼

srbarrios:
Example:
https://ci.suse.de/blue/organizations/jenkins/uyuni-master-dev-acceptance-tests-NUE/detail/uyuni-master-dev-acceptance-tests-NUE/6221/pipeline

Bischoff (approved):

PR #1404, opened 2024-10-02
Move product_version to base level.

maximenoel8:
## What does this PR ? 

After the changes made for sumaform with https://github.com/uyuni-project/sumaform/pull/1713, we can now only declare product_version at the base level and remove it from the individual modules.


Bischoff (approved):
Thank you

To be merged only at same time as corresponding sumaform changes, I assume?

PR #1406, opened 2024-10-03
 add a temporary variable nodeTag to use the correct rake task file name for sle1X minion

maximenoel8:
## What does this PR ?
I align the minionList name with the module names in CI.
Unfortunately, the rake task file are still named with the version sle1X and not sles1X.
I want to move those tags to sles1X too in the testsuite project but it will require more time.
To fix this issue, add a temporary variable nodeTag to use the correct rake task file name for sle1X minion.


Bischoff (approved):
Don't forget to remove the new variable once you have the final fix in the test suite :)

PR #1416, opened 2024-10-07
Kafka initial implementation

szachovy:
Initial implementation of https://github.com/SUSE/spacewalk/issues/24966.

### DIFF
--- a/jenkins_pipelines/scripts/kafka/Dockerfile
+++ b/jenkins_pipelines/scripts/kafka/Dockerfile
@@ -2,19 +2,8 @@ FROM apache/kafka:3.7.1
 
 USER root
 
-RUN \
-  apk \
-    update \
-  && \
-  apk \
-    add \
-      git=2.43.5-r0 \
-      python3=3.11.10-r0 \
-      python3-dev=3.11.10-r0 \
-      py3-pip=23.3.1-r0 \
-      gcc=13.2.1_git20231014-r0 \
-      g++=13.2.1_git20231014-r0 \
-      librdkafka-dev=2.3.0-r1
+RUN apk update && \
+  apk add git=2.43.5-r0 python3=3.11.10-r0 python3-dev=3.11.10-r0 py3-pip=23.3.1-r0 gcc=13.2.1_git20231014-r0 g++=13.2.1_git20231014-r0 librdkafka-dev=2.3.0-r1
 
 USER appuser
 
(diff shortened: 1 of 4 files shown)


jenkins_pipelines/scripts/kafka/Dockerfile
```diff
@@ -0,0 +1,58 @@
...
+RUN \
+  python3 \
+    -m \
+      venv \
+      "/home/appuser/venv" \
+  && \
+  . \
+    "/home/appuser/venv/bin/activate" \
+  && \
+  pip \
```
Bischoff, line 34: wondering if we could use a SUSE package rather than using pip

jenkins_pipelines/scripts/kafka/README.md
```diff
@@ -0,0 +1,62 @@
+# Kafka automation concept
+
+Messaging system to speed up business processes via services API.
```
Bischoff, line 3: what does that mean ?

jenkins_pipelines/scripts/kafka/producer.py
```diff
@@ -0,0 +1,25 @@
+
```
Bischoff, line 1: extra blank line

Bischoff (commented):
see remarks

jenkins_pipelines/scripts/kafka/consumer.py
```diff
@@ -0,0 +1,205 @@
...
+                os.rename('custom_repositories.json', f'custom_repositories_{build_number}.json')
+                return build_number
+            except KeyError:
+                logging.error(f"Build number {build_number} was not found in the currently running pipelines, latest output: {response['why']}. Please check if someone else is working on the pipeline.")
+                return None
+        if request.status_code == 431:
+            logging.error(f'Request is too big, perhaps too many RRs to be accepted generated big JSON, run pipeline manually using generated custom_repositories.json in {os.getcwd()}')
+        return None
+
+    def pipeline_status(self):
```
NamelessOne91, line 146: Nitpicky, but maybe is worth changing this method name and signature to something like :

```python
def pipeline_enabled(self) -> bool:
```

jenkins_pipelines/scripts/kafka/consumer.py
```diff
@@ -0,0 +1,205 @@
...
+        except git.exc.GitError as stderr:
+            logging.warning(f"Error during git pull on susemanager-ci repository: {stderr}")
+
+    @staticmethod
+    def generate_custom_repositories(incidents: dict):
+        incident_numbers = ','.join(
+            str(incident['incident']['incident_id'])
+            for incident in incidents['data']
+        )
+        subprocess.run(
```
NamelessOne91, line 77: Having set `check=True`, if the JSON generating scripts fails, for whatever reason, this function call will raise a `subprocess.CalledProcessError`

Maybe we should handle the possibility of a failure here, or where the function is called, and at least log some useful info.

jenkins_pipelines/scripts/kafka/Dockerfile
szachovy: There are packages like this [one](https://packagehub.suse.com/packages/python-confluent-kafka/) available, but it would require registering SUSE product in the container and then enabling the SUSE Package Hub extension ([link](https://packagehub.suse.com/how-to-use/)). I would rather do not use SUSE credentials to make it run.

jenkins_pipelines/scripts/kafka/Dockerfile
```diff
@@ -0,0 +1,58 @@
...
+COPY --chown=appuser "consumer.py" "producer.py" "/home/appuser/"
+
+WORKDIR "/home/appuser"
+
+RUN \
+  python3 \
+    -m \
+      venv \
+      "/home/appuser/venv" \
+  && \
```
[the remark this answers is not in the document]
szachovy: In the Dockerfile best practises it was to [minimize number of layers](https://stackoverflow.com/questions/47079114/should-i-minimize-the-number-of-docker-layers), not sure what is the status of it now. There are a few pros and cons from both approaches, to me it was preferable to keep it like that, but no strong opinion here.

jenkins_pipelines/scripts/kafka/README.md
szachovy: I used to much marketing here, corrected sentence is:
`Messaging system to automate manual tasks with SUSE internal services via API.`

jenkins_pipelines/scripts/kafka/producer.py
szachovy: done

jenkins_pipelines/scripts/kafka/consumer.py
szachovy: done

jenkins_pipelines/scripts/kafka/consumer.py
szachovy: Good catch, done

jenkins_pipelines/scripts/kafka/Dockerfile
Bischoff: ok

Bischoff (commented):
more comments

jenkins_pipelines/scripts/kafka/Dockerfile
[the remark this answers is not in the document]
szachovy: Each RUN instruction creates a new layer in the Docker image ([source](https://www.docker.com/blog/docker-best-practices-choosing-between-run-cmd-and-entrypoint/)), commands are now combined.

ktsamis (approved):

PR #1419, opened 2024-10-08
Create sle maintenance update pipeline nue

maximenoel8:
Related to: https://github.com/SUSE/spacewalk/issues/25447
Depends on: https://gitlab.suse.de/galaxy/infrastructure/-/merge_requests/970

### DIFF
--- a/terracumber_config/tf_files/SUSEManager-4.3-SLE-update-NUE.tf
+++ b/terracumber_config/tf_files/SUSEManager-4.3-SLE-update-NUE.tf
@@ -66,12 +66,12 @@ variable "SCC_PASSWORD" {
 
 variable "GIT_USER" {
   type = string
-  default = null // Not needed for master, as it is public
+  default = null
 }
 
 variable "GIT_PASSWORD" {
   type = string
-  default = null // Not needed for master, as it is public
+  default = null
 }
 
 terraform {
@@ -115,7 +115,7 @@ module "server" {
   base_configuration = module.base.configuration
   name               = "server"
   provider_settings = {
-    mac                = "aa:b2:93:01:03:98"
+    mac                = "aa:b2:93:01:02:61"
     memory             = 16384
     vcpu               = 4
     data_pool          = "ssd"
@@ -150,7 +150,7 @@ module "proxy" {
   base_configuration = module.base.configuration
   name               = "proxy"
   provider_settings  = {
-    mac      = "aa:b2:93:01:03:99"
+    mac      = "aa:b2:93:01:02:62"
     memory   = 4096
   }
   server_configuration = {
@@ -177,7 +177,7 @@ module "sles15sp4_minion" {
   name               = "sles15sp4-minion"
   image              = "sles15sp4o"
   provider_settings  = {
-    mac    = "aa:b2:93:01:03:9a"
+    mac    = "aa:b2:93:01:02:63"
     vcpu   = 2
     memory = 2048
   }
@@ -192,7 +192,7 @@ module "controller" {
   base_configuration = module.base.configuration
   name               = "controller"
   provider_settings = {
-    mac                = "aa:b2:93:01:03:97"
+    mac                = "aa:b2:93:01:02:60"
     memory             = 16384
     vcpu               = 8
   }
--- a/terracumber_config/tf_files/SUSEManager-4.3-SLE-update-NUE.tf
+++ b/terracumber_config/tf_files/SUSEManager-4.3-SLE-update-NUE.tf
@@ -1,0 +1,216 @@
+// Mandatory variables for terracumber
+variable "URL_PREFIX" {
+  type = string
+  default = "https://ci.suse.de/view/Manager/view/Manager-4.3/job/manager-4.3-qe-sle-update-NUE"
+}
+
+// Not really used as this is for --runall parameter, and we run cucumber step by step
+variable "CUCUMBER_COMMAND" {
+  type = string
+  default = "export PRODUCT='SUSE-Manager' && run-testsuite"
+}
+
+variable "CUCUMBER_GITREPO" {
+  type = string
+  default = "https://github.com/SUSE/spacewalk.git"
+}
+
+variable "CUCUMBER_BRANCH" {
+  type = string
+  default = "Manager-4.3"
+}
+
+variable "CUCUMBER_RESULTS" {
+  type = string
+  default = "/root/spacewalk/testsuite"
+}
+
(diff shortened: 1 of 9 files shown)


terracumber_config/tf_files/SUSEManager-4.3-SLE-update-NUE.tf
```diff
@@ -0,0 +1,216 @@
...
+  type = string
+}
+
+variable "SCC_PASSWORD" {
+  type = string
+}
+
+variable "GIT_USER" {
+  type = string
+  default = null // Not needed for master, as it is public
```
srbarrios, line 69: That message is wrong, we are in spacewalk.

terracumber_config/tf_files/SUSEManager-5.0-SLE-update-NUE.tf
```diff
@@ -0,0 +1,219 @@
...
+  type = string
+}
+
+variable "SCC_PASSWORD" {
+  type = string
+}
+
+variable "GIT_USER" {
+  type = string
+  default = null // Not needed for master, as it is public
```
srbarrios, line 69: Wrong message

terracumber_config/tf_files/SUSEManager-5.0-SLE-update-NUE.tf
```diff
@@ -0,0 +1,219 @@
...
+  runtime               = "podman"
+  // Temporary workaround to see if we pass proxy stage. Also needs to be updated on next MU
+  container_repository  = "registry.suse.de/suse/maintenance/35237/suse_sle-15-sp6_update_products_manager50_update_containerfile"
+  container_tag         = "latest"
+  server_mounted_mirror = "minima-mirror-ci-bv.mgr.suse.de"
+
+  auto_accept                    = false
+  monitored                      = true
+  disable_firewall               = false
+  allow_postgres_connections     = false
```
srbarrios, line 135: Do we really need to set to false all these variables?
if so, maybe we can consider change the default values of them.

srbarrios (approved):

terracumber_config/tf_files/SUSEManager-4.3-SLE-update-NUE.tf
maximenoel8: Correct, it's a copy / paste from the other maintenance sle update. I will update both.

terracumber_config/tf_files/SUSEManager-5.0-SLE-update-NUE.tf
maximenoel8: I can double check the default value in sumaform. Overall in BV / sle maintenance update pipeline we are exhaustive on those parameters. I suppose it's to avoid been impacted by default value change in sumaform.

terracumber_config/tf_files/SUSEManager-5.0-SLE-update-NUE.tf
```diff
@@ -0,0 +1,219 @@
...
+  }
+}
+
+module "server_containerized" {
+  source             = "./modules/server_containerized"
+  base_configuration = module.base.configuration
+  name               = "server"
+  image              = "slemicro55o"
+  provider_settings = {
+    mac                = "aa:b2:93:01:03:9c"
```
Bischoff, line 119: same as for 4.3

terracumber_config/tf_files/SUSEManager-5.0-SLE-update-NUE.tf
```diff
@@ -0,0 +1,219 @@
...
+
+}
+
+module "proxy_containerized" {
+  source             = "./modules/proxy_containerized"
+  base_configuration = module.base.configuration
+  name               = "proxy"
+  image              = "slemicro55o"
+  provider_settings  = {
+    mac                = "aa:b2:93:01:03:9d"
```
Bischoff, line 157: same as for 4.3

terracumber_config/tf_files/SUSEManager-5.0-SLE-update-NUE.tf
```diff
@@ -0,0 +1,219 @@
...
+
+}
+
+module "sles15sp6_minion" {
+  source             = "./modules/minion"
+  base_configuration = module.base.configuration
+  name               = "sles15sp6-minion"
+  image              = "sles15sp6o"
+  provider_settings  = {
+    mac    = "aa:b2:93:01:03:9e"
```
Bischoff, line 182: same as for 4.3

terracumber_config/tf_files/SUSEManager-5.0-SLE-update-NUE.tf
```diff
@@ -0,0 +1,219 @@
...
+  ssh_key_path            = "./salt/controller/id_rsa.pub"
+
+}
+
+module "controller" {
+  source             = "./modules/controller"
+  base_configuration = module.base.configuration
+  name               = "controller"
+  provider_settings = {
+    mac                = "aa:b2:93:01:03:9b"
```
Bischoff, line 198: same as for 4.3

Bischoff (changes requested):
offset by 1 in the MACs

alternatively, make use of new network where there's a lot more room

terracumber_config/tf_files/SUSEManager-5.0-SLE-update-NUE.tf
srbarrios: I'd consider QE squad the main user of this tool, and so, we better align the default values with our default use cases (CI and BV)

maximenoel8: I checked on sumaform side, we need those values. Changing the default is a bit risky, we don't know how other project are using sumaform. It also helps to see quickly without opening sumaform how the server is set. I would like to keep those value as is. We maybe think about changing the default if we create a template + variable files. I tried to remove all the useless parameter ( like server_configuration ) that are already not needed. Let's see for the default.

Bischoff (approved):
LGTM now, thanks for the fixes

terracumber_config/tf_files/SUSEManager-5.0-SLE-update-NUE.tf
srbarrios: Ok, if you both prefer setting all the variables in these environments.

📢 Still, out of scope of this PR, but I think that QE own the main need of sumaform, and we must adapt the default values to our default needs.

ktsamis (approved):

PR #1422, opened 2024-10-08
Changing the Redis instance

srbarrios:
We will start using our own Redis instance.
This user it's only able to push data from 19269 port.
This data is totally ephemeral, as it will be replaced every week.
This data will not contain any information that could be considered a risk, just a map of which tests triggers each product code line.

---

⚠️ Still, I would like to find a way to hide these credentials, probably through infrastructure repository?

Like we do here:
https://gitlab.suse.de/galaxy/infrastructure/-/blob/master/srv/salt/sumaform-node/credentials
https://gitlab.suse.de/galaxy/infrastructure/-/blob/master/srv/pillar/sumaform-node.sls


srbarrios:
OK, we definitely must go through infrastructure repo, as this repo is public.

PR #1429, opened 2024-10-11
Move acceptance tests at PRs to NUE

jordimassaguerpla:
Running this jenkins pipeline at the same place we have the openbuildservice should make the process faster and less complex.


maximenoel8 (approved):
LGTM thank you !

jenkins_pipelines/environments/uyuni-prs-ci-tests
```diff
@@ -1,10 +1,10 @@
 #!/usr/bin/env groovy
 
-node('pull-request-test') {
+node('pull-request-test-nue') {
```
srbarrios, line 3: We don't care too much where this node is located and this is only used to trigger the pipeline. But.. see below.

jenkins_pipelines/environments/uyuni-prs-ci-tests
```diff
@@ -35,6 +35,14 @@ node('pull-request-test') {
         first_env = 1;
         last_env = 10;
         email_to = params.email_to;
+        if (platform_localisation == "NUE") {
+          domain = "mgr.suse.de"
```
srbarrios, line 39: Personally, and really just a matter of taste, I would have both domains as constants in `pipeline-pull-request-envs.groovy`. 
Then use the `platform_localisation` passed as string parameter in the pipeline to calculate which domain to use inside `pipeline-pull-request-envs.groovy`.

The reasoning about it, is that those domains might change over the years and we might have multiple pipelines using the `pipeline-pull-request-envs.groovy`, so it looks more maintainable fmpov.

srbarrios:
Unrelated topic:

image


It will be good to filter certain terraform files, so the checks did not fail.
For instance, the templates.

jordimassaguerpla:
Tests pass now!!

PR #1430, opened 2024-10-11
Verify previous Terraform state can be reused

NamelessOne91:
Related https://github.com/SUSE/spacewalk/issues/25242

Introduces additional checks when "use_previous_terraform_state"is checked,  with the aim to validate a previous Terraform state is indeed present and can be safely reused.

### DIFF
--- a/jenkins_pipelines/environments/common/pipeline-build-validation.groovy
+++ b/jenkins_pipelines/environments/common/pipeline-build-validation.groovy
@@ -28,7 +28,7 @@ def run(params) {
         }
         try {
             stage('Clone terracumber, susemanager-ci and sumaform') {
-                // Create a directory for  to place the directory with the build results (if it does not exist)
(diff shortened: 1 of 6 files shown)


srbarrios:
Looks a nice improvement.
As suggestion, I would also try to use virsh to collect the actual resources status (for instance VMs and disks created), as the tfstate it might actually don't match the real state.
And if I'm guessing right, terraform plan only compares the information from tfstate and main.tf, without directly checking the real deployed resources.

NamelessOne91:
> And if I'm guessing right, terraform plan only compares the information from tfstate and main.tf, without directly checking the real deployed resources.

IIRC plan does automatically a refresh to check differences between deployed resources and what's in the state file ... but I may not RC 😄 
So, it seems there's an option to force it anyway.

I added it and refactored the code a bit to be more robust. 
Still needs to be tested properly, as I have a few doubts it will work and do what we expect.

PR #1433, opened 2024-10-15
TEST-FIX: Disable installing salt bundle on Salt Shaker classic leap for

m-czernek:


m-czernek:
Pipeline currently broken, merging first to test a larger fix...

PR #1437, opened 2024-10-16
Add 5.0 registry for both SLE envs

ktsamis:
We should be using the latest available released container from the registry for SLE, and we can overwrite it as we want with the new jenkins parameter. This registry currently works and deploys. The previous was not until the MU was ready and would not after it was released. So this with the latest tag and the server image from mgradm works now. Tested on NUE deployment


srbarrios (approved):

maximenoel8:
Thanks!

PR #1450, opened 2024-10-22
Changing the container repository path for mgrpxy

srbarrios:
Card https://github.com/SUSE/spacewalk/issues/25577

With the new mgrpxy version ( > 0.1.21 ), we don't need the /suse/manager/5.0/x86_64 anymore for proxy_containerized.

### DIFF
--- a/terracumber_config/tf_files/SUSEManager-5.0-build-validation-NUE.tf
+++ b/terracumber_config/tf_files/SUSEManager-5.0-build-validation-NUE.tf
@@ -218,8 +218,7 @@ module "proxy_containerized" {
     memory             = 4096
   }
   runtime                   = "podman"
-  // The proxy is not appending this path via product code, we need to do it in our infra
-  container_repository = "${var.CONTAINER_REPOSITORY}/suse/manager/5.0/x86_64"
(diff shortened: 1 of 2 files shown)


srbarrios:
Take into account that on SLES Updates, we might still using an older version if we run any pipeline before Nov. 10th, when we release the new MU for mgrpxy.

So, let's do it in two phases, and change it there only after 0.1.23 is released.

Bischoff (approved):

nodeg:
> With the new mgrpxy version ( > 0.1.21 ), we don't need the /suse/manager/5.0/x86_64 anymore for proxy_containerized.

According to e.g. https://build.suse.de/projects/Devel:Galaxy:Manager:5.0/packages/uyuni-tools/files/uyuni-tools.changes?expand=1, I was not able to find that reflected in the release notes/changes.

aaannz:
I think it's `Use the same logic for image computation in mgradm and mgrpxy` [bsc#1228026](https://bugzilla.suse.com/show_bug.cgi?id=1228026)

NamelessOne91 (approved):

PR #1454, opened 2024-10-25
Trigger rubocop only if ruby files has been changed

szachovy:
Request from the discussion: https://suse.slack.com/archives/C02D12TNYLS/p1729753176627919

### DIFF
--- a/jenkins_pipelines/manager_prs/Jenkinsfile_prs_ruby_rubocop
+++ b/jenkins_pipelines/manager_prs/Jenkinsfile_prs_ruby_rubocop
@@ -1,7 +1,7 @@
 #!/usr/bin/env groovy
 
 properties([
-        buildDiscarder(logRotator(numToKeepStr: '500'))
+    buildDiscarder(logRotator(numToKeepStr: '500'))
 ])
 
 pipeline {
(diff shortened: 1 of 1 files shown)


jenkins_pipelines/manager_prs/Jenkinsfile_prs_ruby_rubocop
```diff
@@ -51,12 +56,17 @@ pipeline {
             }
         }
         stage('Run tests against PR') {
+            when {
+                anyOf {
+                    changeset "**/*.rb"
```
srbarrios, line 61: That will not take into account Rakefile

szachovy: In GH actions we have it like that: https://github.com/SUSE/spacewalk/blob/6fbfe9b94e87a0c1b0d025c3a266fc84c396b3ba/.github/workflows/rubocop.yml#L5

I followed this convention, in the recent commit also for `yml` files. I have no problem to add `Gemfile` and `Rakefile` if needed.

srbarrios: Good catch! I think we should include both Gemfile and Rakefile in both GH Actions and Jenkinsfile. As few days ago, we had an issue in Rakefile if I recall correctly. Would you mind to take care of both changes please?

szachovy: Done

maximenoel8 (approved):

ktsamis (approved):

srbarrios (commented):
Thank you!

PR #1459, opened 2024-11-07
Temporary incr. of main disk until we implement a partition for /var/lib/containers/storage/overlay

srbarrios:
In our server containerized, we don't have a separated partition for the new overlay, located in `/var/lib/containers/storage/overlay`, that's causing a full disk issue in `/var`.

For now, I'm fixing it by increasing the main disk size, which automatically will give more space to `/var` partition. [But ideally we should implement a new partition for the overlay through sumaform.](https://github.com/SUSE/spacewalk/issues/25719)

### DIFF
--- a/terracumber_config/tf_files/SUSEManager-5.0-build-validation-NUE.tf
+++ b/terracumber_config/tf_files/SUSEManager-5.0-build-validation-NUE.tf
@@ -183,7 +183,7 @@ module "server_containerized" {
   }
   runtime = "podman"
   container_repository  = var.CONTAINER_REPOSITORY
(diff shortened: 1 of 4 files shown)


maximenoel8 (approved):
Thanks !

szachovy (approved):
If it's only about BV then approved, but there are files where this value is even lower than 40 (Uyuni-Master)

srbarrios:
> If it's only about BV then approved, but there are files where this value is even lower than 40 (Uyuni-Master)

Yup, there is fine, because we don't need to synchronize all the products.

PR #1508, opened 2025-01-10
Improve bv aws

maximenoel8:
## Context
Currently, this pipeline is only deploying the paygo environment without configuring the server

## What does this PR
Add the options to configure the server.

### DIFF
--- a/terracumber_config/tf_files/local_mirror.tf
+++ b/terracumber_config/tf_files/local_mirror.tf
@@ -93,7 +93,7 @@ terraform {
 
 
 provider "libvirt" {
-  uri = "qemu+ssh://root@suma-05.mgr.suse.de/system"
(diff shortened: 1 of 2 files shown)


terracumber_config/tf_files/local_mirror.tf
```diff
@@ -93,7 +93,7 @@ terraform {
 
 
 provider "libvirt" {
-  uri = "qemu+tcp://suma-05.mgr.suse.de/system"
+  uri = "qemu+ssh://root@suma-05.mgr.suse.de/system"
```
NamelessOne91, line 96: I think this can now be switched back to TCP and it should work. We have an idea of why it was not.

That said: is using SSH something that can lead to further issues down the way ?

PR #1529, opened 2025-02-07
Beta client tools are not ready yet

Bischoff:
Disable client tools that are not enabled yet in HEAD.

See https://github.com/uyuni-project/uyuni/pull/9743 : avoid starting to synchronize the "Manager beta client tools" that appear in the product tree, but are in fact the beta client tools for 5.0.

## IMPORTANT NOTE:

In fact, this setting does not seem to work. We had:
```
terracumber_config/tf_files/SUSEManager-Head-NUE.tf:      beta_enabled = true
```

but on the controller:
```
suma-ci-head-controller:~ # env | grep BETA
BETA_ENABLED=False
```

There is probably some bug in sumaform and/or terracumber, that should be fixed before we enable the beta client tools...

### DIFF
--- a/terracumber_config/tf_files/SUSEManager-Head-NUE.tf
+++ b/terracumber_config/tf_files/SUSEManager-Head-NUE.tf
@@ -144,7 +144,7 @@ module "cucumber_testsuite" {
       runtime = "podman"
       container_repository = "registry.suse.de/devel/galaxy/manager/head/containerfile"
       container_tag = "latest"
-      beta_enabled = true
+      beta_enabled = false
     }
     proxy_containerized = {
       provider_settings = {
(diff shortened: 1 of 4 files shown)


deneb-alpha (approved):
LGTM. Please note, I didn't dig if the beta flag was enabled in other tf files not part of this PR

srbarrios (approved):

juliogonzalez (approved):
LGTM. If we're to test on head, this will at least still provide the old client tools which (for the moment) are valid for some "exploratory testing" and should work (again, for the moment, as soon we'll need to switch to the new rebranded client tools for both beta and non-beta)

NamelessOne91 (approved):

Bischoff:
> LGTM. If we're to test on head, this will at least still provide the old client tools which (for the moment) are valid for some "exploratory testing" and should work (again, for the moment, as soon we'll need to switch to the new rebranded client tools for both beta and non-beta)

* If we want to use the old beta client tools, I have to:
    a) NOT merge this PR, at least not in its current form
    b) fix the sumaform/terracumber bug mentioned in the description

* If we want to NOT use the old beta client tools, I have to:
    Stop enabling them from the products list (hoping it's even possible, I fear it's triggered as part of "recommended")

As a side note, the CI is not really for exploratory testing, but the reference environments are.

PR #1560, opened 2025-02-21
Continuous BV: use same order as all other BVs

Bischoff:
Follow organization defined in https://gitlab.suse.de/e_bischoff/infrastructure/-/blob/master/srv/salt/dhcpd-server/README.md?ref_type=heads#bv-testsuite-in-prg-2-and-slc-1


maximenoel8:
I don't understand why removing the order.

Bischoff:
> I don't understand why removing the order.

If we have same order in all test suites, it makes maintenance easier.

BTW, I would also encourage to have same order in all main.tf files...


> Plus taking the risk of introducing errors on working deployment and spending time on something that's working.

True, we could have avoided that by following the documented order in the first place.

PR #1562, opened 2025-02-24
dev-acceptance-tests-code-coverage: use wget to download and extract from tar.gz in the server directly

szachovy:


nodeg (approved):

srbarrios (approved):
At first glance lgtm

PR #1567, opened 2025-02-27
Remove  proxy registry path extension in 5.0 SLE updates

NamelessOne91:
Path extension for Proxy should be no longer needed.


Bischoff (approved):
I saws it the other day, but forgot to do the PR :(

szachovy (approved):

PR #1579, opened 2025-03-06
Add SLES 15 SP6 as potential base OS

Bischoff:
## What does this PR change?

Add SLES 15 SP6 as base OS for SUSE Manager 5.0

## Links

See https://github.com/SUSE/spacewalk/pull/26622

### DIFF
--- a/jenkins_pipelines/environments/manager-5.0-qe-alternative-base-OS-NUE
+++ b/jenkins_pipelines/environments/manager-5.0-qe-alternative-base-OS-NUE
@@ -1,0 +1,51 @@
+#!/usr/bin/env groovy
+
+node('sumaform-cucumber') {
+    def minionList = 'sles15sp6_minion'
+    properties([
+        buildDiscarder(logRotator(numToKeepStr: '5', artifactNumToKeepStr: '3')),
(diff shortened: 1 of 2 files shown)


terracumber_config/tf_files/SUSEManager-5.0-alternate-base-OS-NUE.tf
```diff
@@ -0,0 +1,220 @@
+// Mandatory variables for terracumber
+variable "URL_PREFIX" {
+  type = string
+  default = "https://ci.suse.de/view/Manager/view/Manager-5.0/job/manager-5.0-qe-sle-update-NUE"
```
ktsamis, line 4: Maybe we should rename this to differentiate from the other? This already exists

ktsamis (changes requested):
Clarification for the url

terracumber_config/tf_files/SUSEManager-5.0-alternate-base-OS-NUE.tf
Bischoff: Done.

ktsamis (approved):

PR #1581, opened 2025-03-06
Use ed_25519 private keys for DHCP-DNS setup

NamelessOne91:


Bischoff (approved):
Approving, but there's still one failing check

PR #1600, opened 2025-03-20
set 5.1 product version for 51 BV and continuous bv

maximenoel8:
## What does this PR
Set 5.1 BV and continuous BV to use 5.1 released product.
Depends on: https://github.com/uyuni-project/sumaform/pull/1803


NamelessOne91 (dismissed):

nodeg:
Do we really want to mix the continuous BV with the 5.1 BV? What is the reason behind this, Maxime?

jenkins_pipelines/environments/manager-qe-continuous-build-validation-NUE
```diff
@@ -34,13 +34,13 @@ node('sumaform-cucumber') {
             // Temporary: should move to uyuni-project
             string(name: 'terracumber_gitrepo', defaultValue: 'https://github.com/uyuni-project/terracumber.git', description: 'Terracumber Git Repository'),
             string(name: 'terracumber_ref', defaultValue: 'master', description: 'Terracumber Git ref (branch, tag...)'),
-            string(name: 'product_version', defaultValue: 'head', description: 'Server product version'),
```
srbarrios, line 37: I would prefer to keep using head or uyuni master in our continuous validation.

maximenoel8: It was decided not to use uyuni for BV (time consuming to stabilize).

maximenoel8:
> Do we really want to mix the continuous BV with the 5.1 BV? What is the reason behind this, Maxime?

Currently continuous BV is not really used except by me, changing the default value to something that will be stable will be more useful than having something that is not even stable on CI. It's always possible to change it back to head by just changing the the product version and container path.

maximenoel8:
But let's be honest, there is nothing continuous in this BV :-). 100% fail on reposync requiring a manual intervention.

PR #1623, opened 2025-04-11
Use images repo for SL Micro instead of standard

ktsamis:
Use images repo for SL Micro instead of standard


nodeg:
Do we have any implications on this change? I see less packages in the `images` x86 folder comparted to the `standard` one. Ping @deneb-alpha @raulillo82.

deneb-alpha (approved):
Clarified for this MU round.

This can't be a "final" setup, we will need some additional review and thinking. But, at least for this MU round we are almost covered.

PR #1657, opened 2025-05-30
Update sumaform version only when deploying.

maximenoel8:
## What does this PR do?

This PR ensures that `sumaform` is only updated during the deployment phase, rather than on every pipeline run.

### Why is this needed?

Currently, `sumaform` is updated regardless of whether a deployment is happening. This behavior is unnecessary and potentially harmful:

- It introduces the risk of breaking the BV cleaning pipeline if a new `sumaform` version is incompatible with the existing deployment.
- It undermines the stability of persistent environments by unintentionally modifying the `sumaform` workspace.

### What changes?

With this update, `sumaform` will only be updated when a deployment is actually triggered. This keeps the `sumaform` version and workspace consistent and compatible with the existing deployment when environments are reused.

### DIFF
--- a/jenkins_pipelines/environments/common/pipeline-build-validation.groovy
+++ b/jenkins_pipelines/environments/common/pipeline-build-validation.groovy
@@ -34,13 +34,17 @@ def run(params) {
             env.common_params = "${env.common_params} --parallelism ${params.terraform_parallelism}"
         }
         try {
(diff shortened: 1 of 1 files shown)


nodeg:
I like the idea.

NamelessOne91 (approved):
LGTM, thanks !

PR #1664, opened 2025-06-11
Commenting server_http_proxy except in Head

srbarrios:
In order to mitigate our infrastructure issue, we want to stop using the HTTP Proxy in most of our test suites. We will keep it only on Head.

Related Slack thread: https://suse.slack.com/archives/C02EWMAB7QV/p1749637044417959


Bischoff (approved):
Let's go for this, even if it should not make much difference for machines using NFS-mounted mirrors.

PR #1695, opened 2025-07-18
Test: will "tab crashed" disappear with more memory?

Bischoff:
Test if even more memory on controller helps


Bischoff:
let's just test on my branch, and redo a more complete PR if that worked

PR #1698, opened 2025-07-25
Remove openeuler as an option from pipelines

ktsamis:
Remove resources not used, openeuler is not an option anymore and it's causing problems with the resource-cleaner in terracumber.


srbarrios:
thank you

PR #1700, opened 2025-07-25
Rename s390 guest in continuous BV to use 51 name

ktsamis:
Rename s390 guest in continuous BV to use 51 name instead of 50. The MACs are correct


Bischoff (approved):
Yeah, the problem is that continuous BV is not attached to a given product version...

We should infer the s390 name from `var.PRODUCT_VERSION` - ping @maximenoel8 .

PR #1703, opened 2025-07-29
Do not overwrite server and proxy for 50 in json script

ktsamis:
Do not overwrite server and proxy for 50 in json script.

The timeout is to try and speed the execution.

Added 4.3 LTS server and proxy (server confirmed, proxy not confirmed) and web scripting for 5.0

Tested with all versions with some SLE updates and none throw any errors and the 5.0 is not overwritten.


maximenoel8 (approved):
LGTM. Thanks for the added documentation.

PR #1720, opened 2025-08-21
Add BV and SLE MU based on SLES as base OS

Bischoff:
Add BV and SLE MU based on SLES as base OS

See also https://gitlab.suse.de/galaxy/infrastructure/-/merge_requests/1076


srbarrios (approved):

ktsamis (approved):
It looks fine to me but it's difficult with so many similar names and files with small differences. I checked the names and tf files and urls and images and they match from what I see if I didn't miss anything, I didn't check the MACs but we will see if there is any error when we deploy. Frankly I would merge this and run a deployment of all at the same time one by one and see if there is any deployment errors, I think that's the best way to review this and the infra PR

PR #1758, opened 2025-09-30
Uyuni AWS CI: increase server volume size

NamelessOne91:


maximenoel8 (approved):
LGTM. Don't forget to add reviewers so the people can see it !

NamelessOne91:
> LGTM. Don't forget to add reviewers so the people can see it !

Thanks for the review.
I left this in the back because I no longer had troubles with disk space in Uyuni AWS CI.

PR #1851, opened 2026-01-14
[Draft] Small improvement on sanity checks per MI repository provided on the autogenerated JSON

srbarrios:
Related card: https://github.com/SUSE/spacewalk/issues/29401

This just makes a sanity check to find .repo file inside the repo URL, if it doesn't find it it logs an error. But we need to improve this logic to consider Debian-like repos too.

### DIFF
--- a/jenkins_pipelines/scripts/json_generator/maintenance_json_generator.py
+++ b/jenkins_pipelines/scripts/json_generator/maintenance_json_generator.py
@@ -3,6 +3,7 @@ from functools import cache
 import json
 import requests
 import logging
+import re
 
 from ibs_osc_client import IbsOscClient
 from repository_versions import nodes_by_version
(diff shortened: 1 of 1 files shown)


ktsamis:
I think because of the way we iterate on all possible urls and only use the valid ones this might give all the false negatives. Basically we try to see if an MI has a specific client tools repo and if it doesn't we don't use it, which is a valid case since not all MIs have all the suffixes. I think the way this would check, it would log all these non-repos and show them as errors. But I will review this better later on

srbarrios:
Take this PR as an example :)
It was automatically generated by GH Copilot, and I did not test it.

jenkins_pipelines/scripts/json_generator/maintenance_json_generator.py
```diff
@@ -45,11 +46,49 @@ def clean_mi_ids(mi_ids: list[str]) -> set[str]:
...
+    """
     url = f"{IBS_MAINTENANCE_URL_PREFIX}{mi_id}{suffix}"
 
-    res: requests.Response = requests.get(url, timeout=6)
-    return url if res.ok else ""
+    try:
+        res: requests.Response = requests.get(url, timeout=6)
+        res.raise_for_status()
+    except requests.RequestException as e:
+        logging.error(f"Error requesting: {url}; {e}")
```
NamelessOne91: As pointed out by Kostas, this will log all 404s that we get.
Since they tend to be quite a lot, we may want to skip logging here.

NamelessOne91 (commented):
I tried to feed the MI with ID 42236, using 50-sles, as version it correctly recognized these as valid, because of a link to a .repo:

- http://download.suse.de/ibs/SUSE:/Maintenance:/42236/SUSE_Updates_openSUSE-SLE_15.6/
- http://download.suse.de/ibs/SUSE:/Maintenance:/42236/SUSE_Updates_SLE-Micro_5.5_x86_64/

Not a fan of handling this by parsing HTML with regEx, but it seems to work with RPM repos.

The 1st case, where the URL itself is pointing to a .repo file, is probably not going to happen in our setup.

PR #1934, opened 2026-03-12
Add the hypervisor url to the provider so we can call it with virsh d…

maximenoel8:
## What does this PR ?

Add the hypervisor url to provider so we can call it for virsh verification during deployment (currently wait for ip)
Add this link to all

### DIFF
--- a/terracumber_config/tf_files/personal/main_51.tf
+++ b/terracumber_config/tf_files/personal/main_51.tf
@@ -14,8 +14,12 @@ terraform {
   }
 }
 
+locals {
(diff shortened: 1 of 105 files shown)


terracumber_config/tf_files/personal/main_51.tf
```diff
@@ -14,8 +14,12 @@ terraform {
   }
 }
 
+locals {
+  libvirt_uri = "qemu+tcp://${var.ENVIRONMENT_CONFIGURATION[var.ENVIRONMENT].hypervisor}/system"
+}
```
Bischoff, line 19: this introduces a dependancy towards an environment variable -> the main.tf can't work standalone anymore.

Bischoff (changes requested):
we could keep some ideas, but the use of an envvar sounds wrong to me

PR #1975, opened 2026-04-07
Re-enable problematic minions

Bischoff:
All minions fixed by @NamelessOne91 and me in 5.1.3 full BV and 5.2 beta 1 full BV,
excepted Leap 16 (problem with minima).

We did not test in head, but the changes were done in sumaform and on the mirror, so it should work.


nodeg (approved):
Thank you.

PR #1983, opened 2026-04-16
QE: Point HEAD to correct registries and delete minions

Pablogoliva:
Pointing the HEAD RKE2 pipeline to the correct registries and deleting minions in RKE2 pipelines since are not going to be used in the near future.


ktsamis (approved):
I have no idea if the urls are correct but it seems fine to update them and also removing the minions seems fine for now

PR #1987, opened 2026-04-17
QE: Put the whole url of the oci in the name

Pablogoliva:
Put the whole url of the oci in the name instead of split it.


Pablogoliva:
After a discussion, we decided that is not necessary.

PR #2027, opened 2026-05-15
Update test model for MCP

jordimassaguerpla:
gemini-2.5-flash-lite is known to halucinate because it prefers giving a guessed answer than stopping "the flow". It is the wrong model because it makes up hostnames.

gemini-3.1-flash-lite is prefered for this reason and cost is still reasonable.

### DIFF
--- a/jenkins_pipelines/mlm_ai/Jenkinsfile_mlm_51_ai_acceptance_tests
+++ b/jenkins_pipelines/mlm_ai/Jenkinsfile_mlm_51_ai_acceptance_tests
@@ -15,12 +15,12 @@ pipeline {
             name: 'MCP_SERVER_BRANCH',
             defaultValue: 'main',
             description: 'The branch to check out from the mcp-server-uyuni repository.'
-        ),
+        )
         string(
             name: 'AGENT_MODEL',
             defaultValue: 'gemini-3.1-flash-lite',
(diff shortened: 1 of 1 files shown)


jenkins_pipelines/mlm_ai/Jenkinsfile_mlm_51_ai_acceptance_tests
```diff
@@ -15,6 +15,16 @@ pipeline {
             name: 'MCP_SERVER_BRANCH',
             defaultValue: 'main',
             description: 'The branch to check out from the mcp-server-uyuni repository.'
+        ),
+        string(
+            name: 'AGENT_MODEL',
+            defaultValue: 'gemini-3.1-flash-lite',
```
srbarrios, line 21: I though still a preview gemini-3.1-flash-lite-preview

srbarrios (approved):

jenkins_pipelines/mlm_ai/Jenkinsfile_mlm_51_ai_acceptance_tests
jordimassaguerpla: Preview is gemini 3.0 flash ... it is a bit confusing, but I tested it with gemini-3.1-flash-lite

PR #2067, opened 2026-06-15
Add debian 12 to BV list of minions

ktsamis:
Add debian 12 to BV list of minions.

Also commented out ubuntu 26, it's not there yet available to be deployed.


ktsamis:
Closing this because debian 12 is not supported in 5.2 anymore and head is 5.2 also

PR #2074, opened 2026-06-18
Fix BV cleanup pipeline channel and image profile deletion

ktsamis:
### Problem

The BV cleanup pipeline had two issues:

1. **Image profile deletion failing** - The `delete_image_profiles` method tried to delete hardcoded profile names (`suse_os_image_15_sp6`, `suse_os_image_15_sp7`) that don't exist on servers where the retail stage had not been run, causing:
```
xmlrpc.client.Fault: <Fault 10011: 'redstone.xmlrpc.XmlRpcFault: Image Profile not found'>
```

3. **Custom child channels not being deleted** - The `delete_software_channels` method was not deleting certain custom child channels like:
- "Custom Channel for CentOS 7 DVD"
- "Custom Channel for Rocky 8 DVD" 
- "Ubuntu 24.04 Noble AMD64" (label: `ubuntu-2404-noble-amd64`)

#### 1. Image Profile Deletion ([test_environment_cleaner_api.py:122-128](jenkins_pipelines/scripts/test_environment_cleaner/test_environment_cleaner_program/test_environment_cleaner_api.py#L122-L128))
- Changed to list all existing image profiles first
- Only delete profiles that actually exist
- Added error handling to prevent pipeline failure when retail profiles don't exist

#### 2. Channel Deletion ([test_environment_cleaner_api.py:46-73](jenkins_pipelines/scripts/test_environment_cleaner/test_environment_cleaner_program/test_environment_cleaner_api.py#L46-L73))
- **Simplified logic**: Now only deletes custom child channels
- **Keeps**: All parent channels, vendor distribution child channels (SUSE Client Tools, AppStream, etc.)
- **Deletes**:
- Child channels with "custom" in the label (e.g., `custom-centos-7-dvd`, `custom-rocky-8-dvd`)
- Specific non-custom child channels by exact match (currently `ubuntu-2404-noble-amd64`)
- **Protects**: Infrastructure channels containing `proxy`, `monitoring`, `build`, or `terminal`
- **Removed version differentiation**: Uyuni BV is gone; behavior is now consistent across all SUMA versions

### DIFF
--- a/jenkins_pipelines/scripts/test_environment_cleaner/test_environment_cleaner_program/test_environment_cleaner_api.py
+++ b/jenkins_pipelines/scripts/test_environment_cleaner/test_environment_cleaner_program/test_environment_cleaner_api.py
@@ -44,36 +44,41 @@ class ResourceManager:
             self.client.contentmanagement.removeProject(self.session_key, project['label'])
 
     def delete_software_channels(self):
+        """Delete only custom child channels (keeps all parent channels and vendor child channels)."""
+        # Specific child channels to delete by exact label match (use set for O(1) lookup)
+        specific_channels_to_delete = {
+            "ubuntu-2404-noble-amd64"
+        }
+
+        # No need to differentiate between versions (SUMA 4.3, 5.0, 5.1, 5.2 vs Uyuni)
+        # Uyuni BV is now gone, and the behavior should be the same across all versions and products
         channels = self.client.channel.listMyChannels(self.session_key)
-        product_version = self.get_product_version()
-
-
-        if all(version not in product_version for version in ["5.0", "4.3", "5.1"]):
-            for channel in channels:
-                if "custom" in channel['label'] and not any(protected in channel['label'] for protected in self.resources_to_keep):
-                    logger.info(f"Delete custom channel: {channel['label']}")
-                    self.client.channel.software.delete(self.session_key, channel['label'])
-            logging.warning("Delete only custom channels for uyuni")
-            return
 
         for channel in channels:
-            details = self.client.channel.software.getDetails(self.session_key, channel['label'])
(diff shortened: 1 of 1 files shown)


jenkins_pipelines/scripts/test_environment_cleaner/test_environment_cleaner_program/test_environment_cleaner_api.py
```diff
@@ -44,35 +44,36 @@ def delete_config_projects(self):
             self.client.contentmanagement.removeProject(self.session_key, project['label'])
 
     def delete_software_channels(self):
+        """Delete only custom child channels (keeps all parent channels and vendor child channels)."""
+        # Specific child channels to delete by exact label match
+        specific_channels_to_delete = [
```
NamelessOne91: why do we need to delete this specific channel if it is not a custom child one?

NamelessOne91 (approved):
LGTM, just one question on my side

jenkins_pipelines/scripts/test_environment_cleaner/test_environment_cleaner_program/test_environment_cleaner_api.py
ktsamis: because we recreate in in the Non-MU channels in the pipeline which is part of the client stages. So in the next run after the cleanup this will fail to be created because it's still there. This wastes a little time to resync this channel but I think it's a small one and it saves us a red test later that prevents ubuntu actual bootstrap to run because for that minion the channel creation failed earlier. We could fix that in the step of creating that channel if it exists already but that is a lot riskier on the first run of the pipeline in case the channel is not actually created.

maximenoel8 (approved):

PR #2101, opened 2026-07-13
Drop special sumaform branch for 16.1 hypervisors

Bischoff:
Drop special sumaform branch for 16.1 hypervisors

With my apologies


nodeg (commented):
LGTM

PR #2107, opened 2026-07-15
Fix null interpolation of container registry params and missing HCL brace

ktsamis:
Addresses the Copilot review comments raised on #2106, which were valid but out of scope for that pure-rename PR. All three issues pre-exist on `master`, so they are fixed here independently.

### Changes

1. **`pipeline-build-validation.groovy`** — `server_/proxy_container_repository` were initialized with `?: null` and then unconditionally injected via `--inject ..._CONTAINER_REPOSITORY=${value}`. When the parameter is unset, Groovy interpolates the literal string `"null"`, which the tfvars generator treats as a real registry value. Changed to `?: ''` to match the existing `server_container_image` handling.

2. **`pipeline-build-validation-cleanup.groovy`** — same `?: null` issue for `proxy_container_repository`, exported as `TF_VAR_PROXY_CONTAINER_REPOSITORY`. Changed to `?: ''` and quoted the export so an empty value stays empty instead of becoming the literal string `null`.

3. **`.github/workflows/sumaform-validation.yml`** — the `SERVER_ADDITIONAL_REPOS` variable in the Static BV heredoc was missing its closing `}`, producing invalid HCL in the generated `variables.tf` and failing `tofu plan`. Added the missing brace.

No functional change beyond the corrected empty/unset handling.

### DIFF
--- a/.github/workflows/sumaform-validation.yml
+++ b/.github/workflows/sumaform-validation.yml
@@ -177,7 +177,7 @@ jobs:
           variable "SERVER_CONTAINER_REPOSITORY" { default="" }
           variable "PROXY_CONTAINER_REPOSITORY" { default="" }
           variable "SERVER_CONTAINER_IMAGE" { default="" }
-          variable "SERVER_ADDITIONAL_REPOS" { default={} 
+          variable "SERVER_ADDITIONAL_REPOS" { default={} }
           variable "PROXY_ADDITIONAL_REPOS" { default={} }
           variable "BASE_OS" { default="" }
           variable "PRODUCT_VERSION" { default="" }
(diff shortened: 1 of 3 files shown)


Bischoff (approved):

.github/workflows/sumaform-validation.yml
```diff
@@ -177,7 +177,7 @@ jobs:
           variable "SERVER_CONTAINER_REPOSITORY" { default="" }
           variable "PROXY_CONTAINER_REPOSITORY" { default="" }
           variable "SERVER_CONTAINER_IMAGE" { default="" }
-          variable "SERVER_ADDITIONAL_REPOS" { default={} 
+          variable "SERVER_ADDITIONAL_REPOS" { default={} }
```
nodeg, line 180: Nice catch. I wonder why this worked before.

NamelessOne91 (approved):

maximenoel8 (approved):

.github/workflows/sumaform-validation.yml
ktsamis: no idea why it worked before, I had the same question

PR #2127, opened 2026-07-22
Fix pipeline for 5.2 acceptance tests

meaksh:
The current pipeline is causing an error:

```
java.lang.NoSuchMethodError: No such DSL method 'activeChoice' found among steps 
```

This PR arrange this pipeline according to what we currently have for 5.1 and Head, moving `activeChoice` in favor of `extendedChoice`.


ktsamis (approved):
it aligns with 5.1/Head, but heads up: the activeChoice in 5.2 wasn't accidental. It came from https://github.com/SUSE/susemanager-ci/pull/1993 (Maxime, merged May 11), which deliberately moved these checkbox params extendedChoice → activeChoice because the Extended Choice plugin is obsolete. The plan there was "extendedChoice on legacy server, activeChoice on the new Jenkins", and it kept extendedChoice copies under environments/legacy/.

Problem is reality doesn't match that: on the new Jenkins activeChoice throws NoSuchMethodError, so the Active Choices plugin clearly isn't installed there — while extendedChoice (5.1/Head) works fine. So your fix is correct for getting 5.2 green now.

But it quietly reverses #1993's direction, and the build-validation pipelines Maxime converted there (5.2/head BV etc.) still use activeChoice → they'll break the same way on the new Jenkins. Probably worth looping in @maximenoel8 : either the Active Choices plugin needs to go on the new Jenkins, or we standardize on extendedChoice and revert #1993's BV conversions too. Happy to merge yours Pablo as the immediate fix either way.

mcalmer (approved):

PR #2131, opened 2026-07-22
Reflect controller host OS move to openSUSE Leap 16.0

ktsamis:
## What does this PR change?

Moves the test-suite **controller** to openSUSE Leap 16.0 (from Leap 15.6, EOL since April 2026), across all environments.

The controller host OS is defined centrally in sumaform (`modules/controller/main.tf`, hardcoded image), so the migration is global — driven by the companion sumaform PR. This PR handles the susemanager-ci side:

1. **Add `opensuse160o` to every environment's image download list** (48 `images = [...]` lists + the 2 uppercase `IMAGES = [...]` lists in `PR-tfvars/PR-testing-{uyuni,manager43}.tfvars`). The controller's base image must be present in each environment's download list, otherwise its base volume won't exist on libvirt. `opensuse156o` is **kept** (the openSUSE Leap 15.6 minion clients still use it) — `opensuse160o` is added alongside it.
2. **`Uyuni-Master-AWS.tf`** — this environment *explicitly overrides* the controller image; flipped `opensuse156o` → `opensuse160o` (it would otherwise stay pinned to 15.6).
3. **`README.md`** — image matrix: all Controller rows → Leap 16.0.

Not touched:
- `opensuse156o` stays wherever it's used for minion clients / dhcp_dns / mirrors.
- **salt-shaker** environments (their `opensuse156o` is the system-under-test, and their controller comes from the separate `salt_testenv` module — not part of this migration).
- Mirror-only configs (no controller).

Companion PRs:
- sumaform: https://github.com/uyuni-project/sumaform/pull/2230 (controller VM image → Leap 16.0)
- uyuni: https://github.com/uyuni-project/uyuni/pull/12313 (controller-dev container → Leap 16.0)

Links: https://github.com/SUSE/spacewalk/issues/31270

### DIFF
--- a/terracumber_config/tf_files/personal/main_43.tf
+++ b/terracumber_config/tf_files/personal/main_43.tf
@@ -28,7 +28,7 @@ module "cucumber_testsuite" {
   cc_username = var.SCC_USER
   cc_password = var.SCC_PASSWORD
 
-  images = ["rocky8o", "opensuse156o", "sles15sp4o", "ubuntu2204o"]
+  images = ["rocky8o", "opensuse156o", "opensuse160o", "sles15sp4o", "ubuntu2204o"]
 
   ssh_key_path = var.CONTROLLER_PUBLIC_SSH_KEY_PATH
   use_avahi    = false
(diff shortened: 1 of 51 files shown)


terracumber_config/tf_files/personal/main_43.tf
```diff
@@ -28,7 +28,7 @@ module "cucumber_testsuite" {
   cc_username = var.SCC_USER
   cc_password = var.SCC_PASSWORD
 
-  images = ["rocky8o", "opensuse156o", "sles15sp4o", "ubuntu2204o"]
+  images = ["rocky8o", "opensuse156o", "opensuse160o", "sles15sp4o", "ubuntu2204o"]
```
Bischoff, line 31: is there a place where we still need openSUSE 15.6 ?

Bischoff (changes requested):
what about removing 15.6 ?

terracumber_config/tf_files/personal/main_43.tf
ktsamis: yes for minions. For now, I want to only add for now and then we have a separate card for removing 15.6: https://github.com/SUSE/spacewalk/issues/31265 which will be a follow up. As long as the controller is on leap 16. I can close the card about bumping the controller and then later take the other card (way too many things at the same time currently)

Bischoff: Hmmm, I feel it as an update operation ie an addition + a removal.

But as you wish.

Bischoff (approved):
approving but with slight unagreement :)

PR #2143, opened 2026-07-27
Add the 5.2 reference environment job on the new Jenkins

ktsamis:
## What

Adds a 5.2 reference environment job that runs on the new Jenkins
(`jenkins.mgr.suse.de`), equivalent to `manager-5.1-infra-reference-NUE` on
ci.suse.de.

## Changes

- **`terracumber_config/tf_files/MLM-5.2-refenv-PRG.tf`** (new) — copy of
  `MLM-5.1-refenv-NUE.tf` adapted to 5.2: `product_version = "5.2-nightly"`,
  `slmicro62o` for server and proxy, `name_prefix = "mlm-ref-52-"`, the already
  reserved `aa:b2:93:01:02:f0`–`fd` MACs on vlan2509 (`bridge = "br1"`), and
  `CONTROLLER_PUBLIC_SSH_KEY_PATH` / `HYPERVISOR_PRIVATE_SSH_KEY_PATH`
  variables so the new Jenkins can point at the keys in the agent home.
- **`jenkins_pipelines/environments/manager-5.2-infra-reference`** (new) —
  the job definition, testsuite branch `Manager-5.2`, backend `libvirt`,
  `/usr/bin/tofu`, nightly `H 3 * * *` trigger.
- **`jenkins_pipelines/environments/common/pipeline-reference-new.groovy`** —
  the new Jenkins has no `/home/jenkins/.credentials` on its agents, so this
  ports the mechanism already used by `common/pipeline.groovy` (commit
  71900b44): detect `JENKINS_URL`, source the `sumaform-secrets` credential
  from a temporary file and write the ssh key paths into `terraform.tfvars`.
  The ci.suse.de path is unchanged, so the 4.3 / 5.0 / 5.1 / Head / Uyuni
  reference jobs are not affected.

## Notes

- DHCP and DNS entries for `mlm-ref-52-*` already exist, so no infrastructure
  change is required.
- The registry stays plain `registry.suse.de`: the ToTest registry workaround
  applies to `-released` maintenance update deployments, while this is a
  nightly environment.

## Testing

- `tofu validate` passes against the sumaform modules.
- The tf file was diffed against `MLM-5.1-refenv-NUE.tf` to confirm only the
  intended differences.
- The job still has to be created on jenkins.mgr.suse.de and added to the
  By Functionality → Reference view before the first run.

### DIFF
--- a/jenkins_pipelines/environments/common/pipeline-reference-new.groovy
+++ b/jenkins_pipelines/environments/common/pipeline-reference-new.groovy
@@ -9,58 +9,94 @@ def run(params) {
             env.common_params = "${env.common_params} --parallelism ${params.deploy_parallelism}"
         }
 
-        try {
-            stage('Clone terracumber, susemanager-ci and sumaform') {
-                // Create a directory for  to place the directory with the build results (if it does not exist)
-                sh "mkdir -p ${resultdir}"
-                git url: params.terracumber_gitrepo, branch: params.terracumber_ref
-                dir("susemanager-ci") {
-                    checkout scm
-                }
-                // Clone sumaform
-                sh "set +x; source /home/jenkins/.credentials set -x; ./terracumber-cli ${common_params} --gitrepo ${params.sumaform_gitrepo} --gitref ${params.sumaform_ref} --runstep gitsync"
-
-                // Restore Terraform states from artifacts
-                if (params.use_previous_terraform_state) {
-                    copyArtifacts projectName: currentBuild.projectName, selector: specific("${currentBuild.previousBuild.number}")
-                }
+        // The new Jenkins has no /home/jenkins/.credentials on its agents: the same content
+        // is provided by the sumaform-secrets credential and sourced from a temporary file.
+        def isNewJenkins = env.JENKINS_URL?.contains('jenkins.mgr.suse.de')
+        def credInit = isNewJenkins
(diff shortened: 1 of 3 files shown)


maximenoel8 (approved):
It looks correct

maximenoel8:
Do you know why we have reference new ?

maximenoel8:
I didn't analyze too much but could reference be merged with pipeline.groovy ?
What are the main differences ?

ktsamis:
Yeah, `-new` is just a bad name - Michael added it in 06f9b87b (May 2023) when the reference envs moved to the sumaform `testsuite` module. The old `pipeline-reference.groovy` is pure terraform apply, no cucumber at all; the `-new` one also runs core / reference_reposync / proxy / reference_init_clients to actually populate the server. Only the two SLC jobs still use the old one.

On merging with `pipeline.groovy`: I looked, and I would rather not. Different rake targets (`cucumber:reference_reposync` + `<ns>:reference_init_clients` vs `cucumber:reposync` + `parallel:init_clients`), and the reference job deliberately has no sanity check, no secondary, no finishing/report/junit/RRTG, and only mails on deploy failure - it builds a long-lived env, it does not report on a test run. Merging means an `isReference` flag switching off two thirds of a file that 18 acceptance jobs depend on.

What is worth merging is the small bit: the `isNewJenkins` / `credInit` / `withCreds` block is now copy-pasted in 5 files. Card for pulling that into a shared helper: SUSE/spacewalk#31431

Also squashed the PR to one commit. The groovy diff looks bigger than it is - most of it is the body being reindented into the `withCreds {` closure.

PR #34, opened 2019-02-19
PR check pipeline for Uyuni

srbarrios:
Porting the new pipeline to Uyuni, and fixing job trigger to be asynchron.

PR #2174, opened 2026-08-06
Abort cucumber runs that stop producing console output

ktsamis:
## What does this PR change?

A Playwright worker can wedge after a step timeout and never exit. The stage then produces no further output but keeps the build alive until the job-level wall-clock timeout: build [4176](https://ci.suse.de/view/Manager/view/Uyuni/job/uyuni-master-dev-acceptance-tests-podman/4176/) sat in `Core - Initialize clients` for 11h37m before being aborted by hand.

Every cucumber run now sits inside an inactivity timeout, which resets on every console line, so a long stage is unaffected and only a silent one is killed. It covers `pipeline.groovy`, `pipeline-personal.groovy`, `pipeline-rke2.groovy`, `pipeline-pull-request.groovy` and both build validation pipelines, all through the same `withIdleTimeout` helper.

## The numbers

The default is **60 minutes everywhere**. The build validation product sync gets **180**, because it is the one stage measured going silent for longer. Both are overridable per job, with `cucumber_idle_timeout` and `product_sync_idle_timeout`; anything that is not a positive number of minutes falls back to the default.

Longest silence between two console lines, per stage, measured over full runs:

| run | length | worst silence | where |
|---|---|---|---|
| [manager-4.3-qe-build-validation #43](https://jenkins.mgr.suse.de/job/manager-4.3-qe-build-validation/43/) | 11.6h | **62.4 min** | Sync. products and channels |
| [manager-5.1-sles-qe-build-validation #9](https://jenkins.mgr.suse.de/job/manager-5.1-sles-qe-build-validation/9/) | 9.0h, SUCCESS | **31.1 min** | Sync. products and channels |
| [uyuni-master-dev-acceptance-tests-podman #4174](https://ci.suse.de/view/Manager/view/Uyuni/job/uyuni-master-dev-acceptance-tests-podman/4174/) | 7.9h | 26 min | a retrying Capybara step |
| [manager-4.3-dev-acceptance-tests #3584](https://ci.suse.de/view/Manager-qe/job/manager-4.3-dev-acceptance-tests/3584/) | 10.2h | 26 min | a retrying Capybara step |

Every build validation stage other than the product sync stayed under 5 minutes. Output arrives from terracumber in buffered bursts, so those gaps are what the timeout actually sees, whatever the test was doing underneath.

Two limits of the approach, both visible in the data. During the parallel client stages the branches interleave in one console, so a single branch's silence cannot be separated out — the per-stage numbers there are lower bounds. And an idle timeout only catches silence: [manager-5.2-sles-qe-build-validation #34](https://jenkins.mgr.suse.de/job/manager-5.2-sles-qe-build-validation/34/) ran for 67 hours without ever going quiet for more than 5.5 minutes, retrying `Add Activation Keys` in a loop. Catching that needs a wall-clock stage timeout, which is a separate change.

## Scope

Only the cucumber call itself is wrapped: waiting on a `lock`, on `randomWait()` or on a `confirm_before_continue` input happens outside the block, so a stage queued behind another one is never at risk. On trigger the stage is marked aborted, and `Save TF state` and `Get results` still run.

In `Secondary features` the `sh` calls sit inside the timeout closure rather than in their `def statusCode… =` initialisers. Those variables default to `1`, so a run that never gets to assign one cannot be read as a success.

The Jenkins layer is a deliberate first step, not the final home. A killer has to live outside the wedged process: the testsuite's own Ruby watchdog can never fire, because the main thread holds the GVL inside a native Playwright call. Whether this belongs in terracumber or in the Rakefile instead is tracked in SUSE/spacewalk#31590, with the options and their trade-offs.

## How was this tested?

[ktsamis-personal-pipeline #31](https://ci.suse.de/view/Manager/view/Manager-qe/job/ktsamis-personal-pipeline/31/console), on a throwaway branch that is not part of this PR. To keep the run short it lowered the timeout to 2 minutes and replaced `Sanity Check` with 4 minutes of output followed by silence. **The 2 minutes below is that test rig only.**

The noisy phase ran to completion at twice the timeout, because every line resets the clock:

```
[16:07:14.350Z] Timeout set to expire after 2 min 0 sec without activity
[16:07:14.678Z] noisy phase, line 1 of 8
...
[16:10:47.597Z] noisy phase, line 8 of 8
```

The silent phase was killed 2m12s in, and the `finally` block still ran:

```
[16:11:19.863Z] going silent now
[16:13:31.862Z] Cancelling nested steps due to timeout
[16:13:31.863Z] Sending interrupt signal to process
[16:13:41.213Z] script returned exit code 143
[16:13:41.344Z] Archiving artifacts
```

### DIFF
--- a/jenkins_pipelines/environments/common/pipeline-build-validation-aws.groovy
+++ b/jenkins_pipelines/environments/common/pipeline-build-validation-aws.groovy
@@ -13,6 +13,25 @@ def run(params) {
             def build_validation = true
             env.exports = "export BUILD_NUMBER=${BUILD_NUMBER}; export BUILD_VALIDATION=true; export CUCUMBER_PUBLISH_QUIET=true;"
 
+            // Inactivity timeout: kills a cucumber run that has stopped producing output entirely.
+            // Measured on full runs, the worst legitimate silence outside the product sync stays
+            // under 5 minutes, while the sync itself went silent for 62 minutes in
+            // manager-4.3-qe-build-validation #43, so it gets a timeout of its own.
+            // Anything that is not a positive number of minutes falls back to the default.
+            // No 'def': both values are read from clientTestingStages() and clientMigrationStages().
+            def idleTimeoutMinutesOr = { value, int fallback ->
+                def raw = value?.toString()?.trim()
+                raw?.isInteger() && raw.toInteger() > 0 ? raw.toInteger() : fallback
+            }
+            cucumberIdleTimeoutMinutes = idleTimeoutMinutesOr(params.cucumber_idle_timeout, 60)
+            productSyncIdleTimeoutMinutes = idleTimeoutMinutesOr(params.product_sync_idle_timeout, 180)
+            withIdleTimeoutOf = { int minutes, Closure body ->
+                timeout(activity: true, time: minutes, unit: 'MINUTES') { body() }
+            }
+            withIdleTimeout = { Closure body ->
+                withIdleTimeoutOf(cucumberIdleTimeoutMinutes, body)
+            }
+
             def ssh_option = '-o StrictHostKeyChecking=no -o ConnectTimeout=7200 -o ServerAliveInterval=60'
             String server_ami = params.server_ami ?: ""
             String proxy_ami = params.proxy_ami ?: ""
@@ -212,7 +231,9 @@ def run(params) {
 
                 stage('Run core features') {
                     if (params.must_run_core && (deployed || !params.must_deploy)) {
-                        sh "./terracumber-cli ${common_params} --logfile ${resultdirbuild}/testsuite.log --runstep cucumber --cucumber-cmd '${env.exports} cd /root/spacewalk/testsuite; rake cucumber:build_validation_core'"
+                        withIdleTimeout {
+                            sh "./terracumber-cli ${common_params} --logfile ${resultdirbuild}/testsuite.log --runstep cucumber --cucumber-cmd '${env.exports} cd /root/spacewalk/testsuite; rake cucumber:build_validation_core'"
+                        }
                     }
                 }
 
@@ -220,7 +241,10 @@ def run(params) {
                     if (params.must_sync && (deployed || !params.must_deploy)) {
                         // Get minion list from terraform state list command
                         def nodesHandler = getNodesHandler()
-                        def res_sync_products = sh(script: "./terracumber-cli ${common_params} --logfile ${resultdirbuild}/testsuite.log --runstep cucumber --cucumber-cmd 'unset ${nodesHandler.envVariableListToDisable.join(' ')}; ${env.exports} cd /root/spacewalk/testsuite; rake cucumber:build_validation_reposync'", returnStatus: true)
+                        def res_sync_products = 1
+                        withIdleTimeoutOf(productSyncIdleTimeoutMinutes) {
+                            res_sync_products = sh(script: "./terracumber-cli ${common_params} --logfile ${resultdirbuild}/testsuite.log --runstep cucumber --cucumber-cmd 'unset ${nodesHandler.envVariableListToDisable.join(' ')}; ${env.exports} cd /root/spacewalk/testsuite; rake cucumber:build_validation_reposync'", returnStatus: true)
+                        }
                         echo "Custom channels and MU repositories synchronization status code: ${res_sync_products}"
                         sh "exit ${res_sync_products}"
                     }
@@ -234,7 +258,10 @@ def run(params) {
                             input 'Press any key to start adding Maintenance Update repositories'
                         }
                         echo 'Add custom channels and MU repositories'
-                        def res_mu_repos = sh(script: "./terracumber-cli ${common_params} --logfile ${resultdirbuild}/testsuite.log --runstep cucumber --cucumber-cmd '${env.exports} cd /root/spacewalk/testsuite; rake cucumber:build_validation_add_maintenance_update_repositories_proxy'", returnStatus: true)
+                        def res_mu_repos = 1
+                        withIdleTimeout {
+                            res_mu_repos = sh(script: "./terracumber-cli ${common_params} --logfile ${resultdirbuild}/testsuite.log --runstep cucumber --cucumber-cmd '${env.exports} cd /root/spacewalk/testsuite; rake cucumber:build_validation_add_maintenance_update_repositories_proxy'", returnStatus: true)
+                        }
                         echo "Custom channels and MU repositories status code: ${res_mu_repos}"
                         sh "exit ${res_mu_repos}"
                     }
@@ -245,7 +272,10 @@ def run(params) {
                         if (params.confirm_before_continue) {
                             input 'Press any key to start adding activation keys'
                         }
-                        def res_add_keys = sh(script: "./terracumber-cli ${common_params} --logfile ${resultdirbuild}/testsuite.log --runstep cucumber --cucumber-cmd '${env.exports} cd /root/spacewalk/testsuite; rake cucumber:build_validation_add_activation_key_proxy'", returnStatus: true)
+                        def res_add_keys = 1
+                        withIdleTimeout {
+                            res_add_keys = sh(script: "./terracumber-cli ${common_params} --logfile ${resultdirbuild}/testsuite.log --runstep cucumber --cucumber-cmd '${env.exports} cd /root/spacewalk/testsuite; rake cucumber:build_validation_add_activation_key_proxy'", returnStatus: true)
+                        }
                         echo "Add Proxy Activation Key status code: ${res_add_keys}"
                         sh "exit ${res_add_keys}"
                     }
@@ -256,7 +286,10 @@ def run(params) {
                         if (params.confirm_before_continue) {
                             input 'Press any key to start creating the proxy bootstrap repository'
                         }
-                        def res_create_bootstrap_repos = sh(script: "./terracumber-cli ${common_params} --logfile ${resultdirbuild}/testsuite.log --runstep cucumber --cucumber-cmd '${env.exports} cd /root/spacewalk/testsuite; rake cucumber:build_validation_create_bootstrap_repository_proxy'", returnStatus: true)
+                        def res_create_bootstrap_repos = 1
+                        withIdleTimeout {
+                            res_create_bootstrap_repos = sh(script: "./terracumber-cli ${common_params} --logfile ${resultdirbuild}/testsuite.log --runstep cucumber --cucumber-cmd '${env.exports} cd /root/spacewalk/testsuite; rake cucumber:build_validation_create_bootstrap_repository_proxy'", returnStatus: true)
+                        }
                         echo "Create Proxy bootstrap repository status code: ${res_create_bootstrap_repos}"
                         sh "exit ${res_create_bootstrap_repos}"
                     }
@@ -266,7 +299,10 @@ def run(params) {
                         if (params.confirm_before_continue) {
                             input 'Press any key to start bootstraping the Proxy'
                         }
-                        def res_init_proxy = sh(script: "./terracumber-cli ${common_params} --logfile ${resultdirbuild}/testsuite.log --runstep cucumber --cucumber-cmd '${env.exports} cd /root/spacewalk/testsuite; rake cucumber:build_validation_init_proxy'", returnStatus: true)
+                        def res_init_proxy = 1
+                        withIdleTimeout {
+                            res_init_proxy = sh(script: "./terracumber-cli ${common_params} --logfile ${resultdirbuild}/testsuite.log --runstep cucumber --cucumber-cmd '${env.exports} cd /root/spacewalk/testsuite; rake cucumber:build_validation_init_proxy'", returnStatus: true)
+                        }
                         echo "Init Proxy status code: ${res_init_proxy}"
                         sh "exit ${res_init_proxy}"
                     }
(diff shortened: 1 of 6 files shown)


jenkins_pipelines/environments/common/pipeline.groovy
```diff
@@ -80,6 +80,14 @@ def run(params) {
...
                 }
             }
+            // A wedged Playwright worker stops producing output but never exits, so the job-level
+            // wall-clock timeout only fires hours later. This is an inactivity timeout: the clock
+            // resets on every console line, and the longest legitimate silence measured across full
+            // acceptance runs is ~26 minutes (a single retrying Capybara step).
+            def cucumber_idle_timeout = params.cucumber_idle_timeout == null ? 60 : params.cucumber_idle_timeout
+            def withIdleTimeout = { Closure body ->
+                timeout(activity: true, time: cucumber_idle_timeout, unit: 'MINUTES') { body() }
+            }
```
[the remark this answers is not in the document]
ktsamis: Fixed in 9b233578 — the value is now coerced with `(params.cucumber_idle_timeout ?: 60).toString().toInteger()`, so null, an empty string and 0 all fall back to the default.

jenkins_pipelines/environments/common/pipeline-build-validation.groovy
```diff
@@ -58,6 +58,8 @@ def run(params) {
             if (params.deploy_parallelism) {
                 env.common_params = "${env.common_params} --parallelism ${params.deploy_parallelism}"
             }
+            // Inactivity timeout: kills a cucumber run that has stopped producing output entirely.
+            cucumber_idle_timeout = (params.cucumber_idle_timeout ?: 180).toString().toInteger()
```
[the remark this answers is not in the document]
ktsamis: Not changing this — the file already does exactly that: `mgrCreateBootstrapRepo` and `downloadPackagesLock` are assigned without `def` in `run()` and read from `clientTestingStages()`. `cucumber_idle_timeout` is assigned at the top of `run()`, before any stage or helper call, and `runCucumberRakeTarget` is only ever called from within it.

jenkins_pipelines/environments/common/pipeline-personal.groovy
```diff
@@ -149,9 +162,12 @@ def run(params) {
                             tags_list = "export TAGS='${transformed_scopes}'; "
                         }
 
-                        def statusCode1 = sh script: "./terracumber-cli ${common_params} --logfile ${resultdirbuild}/testsuite.log --runstep cucumber --cucumber-cmd '${tags_list}cd /root/spacewalk/testsuite; ${exports} rake cucumber:secondary'", returnStatus: true
-                        def statusCode2 = sh script: "./terracumber-cli ${common_params} --logfile ${resultdirbuild}/testsuite.log --runstep cucumber --cucumber-cmd '${tags_list}cd /root/spacewalk/testsuite; ${exports} rake ${params.rake_namespace}:secondary_parallelizable'", returnStatus: true
-                        def statusCode3 = sh script: "./terracumber-cli ${common_params} --logfile ${resultdirbuild}/testsuite.log --runstep cucumber --cucumber-cmd '${tags_list}cd /root/spacewalk/testsuite; ${exports} rake ${params.rake_namespace}:secondary_finishing'", returnStatus: true
+                        def statusCode1 = 0
```
maximenoel8: value should be 1 failing. If timeout stay 1

jenkins_pipelines/environments/common/pipeline-rke2.groovy
```diff
@@ -104,9 +121,12 @@ def run(params) {
                             tags_list = "export TAGS='${transformed_scopes}'; "
                         }
 
-                        def statusCode1 = sh script: "./terracumber-cli ${common_params} --logfile ${resultdirbuild}/testsuite.log --runstep cucumber --cucumber-cmd '${tags_list}cd /root/spacewalk/testsuite; ${exports} rake cucumber:secondary'", returnStatus: true
-                        def statusCode2 = sh script: "./terracumber-cli ${common_params} --logfile ${resultdirbuild}/testsuite.log --runstep cucumber --cucumber-cmd '${tags_list}cd /root/spacewalk/testsuite; ${exports} rake ${params.rake_namespace}:secondary_parallelizable'", returnStatus: true
-                        def statusCode3 = sh script: "./terracumber-cli ${common_params} --logfile ${resultdirbuild}/testsuite.log --runstep cucumber --cucumber-cmd '${tags_list}cd /root/spacewalk/testsuite; ${exports} rake ${params.rake_namespace}:secondary_finishing'", returnStatus: true
+                        def statusCode1 = 0
```
maximenoel8: same here

jenkins_pipelines/environments/common/pipeline.groovy
```diff
@@ -210,9 +223,12 @@ def run(params) {
                                         .join(' or ')
                                 tags_list = "export TAGS=\"${transformed_scopes}\"; "
                             }
-                            def statusCode1 = sh script: "./terracumber-cli ${common_params} --logfile ${resultdirbuild}/testsuite.log --runstep cucumber --cucumber-cmd '${tags_list}cd /root/spacewalk/testsuite; ${exports} rake cucumber:secondary'", returnStatus: true
-                            def statusCode2 = sh script: "./terracumber-cli ${common_params} --logfile ${resultdirbuild}/testsuite.log --runstep cucumber --cucumber-cmd '${tags_list}cd /root/spacewalk/testsuite; ${exports} rake ${params.rake_namespace}:secondary_parallelizable'", returnStatus: true
-                            def statusCode3 = sh script: "./terracumber-cli ${common_params} --logfile ${resultdirbuild}/testsuite.log --runstep cucumber --cucumber-cmd '${tags_list}cd /root/spacewalk/testsuite; ${exports} rake ${params.rake_namespace}:secondary_finishing'", returnStatus: true
+                            def statusCode1 = 0
```
maximenoel8: same

maximenoel8:
I think 2 minute is way too short. Be careful with this broad approach that may break other stage. 

I like the withIdleTimeout function. Why not use it too in AWS.

Maybe this timeout should be integrated to terracumber instead. So we don't need all those lanes. Or Rakefile.

I try to create a watchdog in testsuite but it can't work.

Here some investigation I did:

[handoff-rrtg-bug031-2026-08-07.md](https://github.com/user-attachments/files/30811932/handoff-rrtg-bug031-2026-08-07.md)

maximenoel8 (changes requested):
The minimum would be the status = 1

I would say we should have this timeout mechanism in terracumber or rake level

jenkins_pipelines/environments/common/pipeline-personal.groovy
ktsamis: Fixed in `30f71cf2` — all `statusCode*` now default to 1. Also applied the same to `pipeline-pull-request.groovy`.

jenkins_pipelines/environments/common/pipeline-rke2.groovy
ktsamis: Fixed in `30f71cf2` — all `statusCode*` now default to 1. Also applied the same to `pipeline-pull-request.groovy`.

jenkins_pipelines/environments/common/pipeline.groovy
ktsamis: Fixed in `30f71cf2` — all `statusCode*` now default to 1. Also applied the same to `pipeline-pull-request.groovy`.

ktsamis:
Two of those are addressed:

- **AWS** — `pipeline-build-validation-aws.groovy` was already covered, but with the raw `timeout(activity: true, ...)` repeated inline. It now uses the `withIdleTimeout` helper at all 22 call sites, same as `pipeline-build-validation.groovy`.
- **2 minutes** — that value is not in this PR. It was a throwaway branch used to make the test run short enough to read. The defaults here are 60 minutes for acceptance, personal, rke2 and pull request, and 180 minutes for both build validation pipelines, all overridable with `cucumber_idle_timeout`. I have reworded the test section so it is not read as the shipped value.

Still on the `status = 1` point: fixed, all `statusCode*` now default to 1.

ktsamis:
On terracumber or rake level: I agree that is probably the better long term home, but I want to keep it here for now. First I want to see this actually working in production, catching idle testsuites and killing them. Once we trust it, we can move it to a different level if we want to.

I opened SUSE/spacewalk#31590 for that decision. It has the options I could think of with pros and cons for each, including your terracumber and Rakefile ideas.

Your handoff doc is in there too. The point that a Ruby thread can never fire while the main thread holds the GVL inside a native Playwright call is the strongest argument for keeping the killer outside the process, so thanks for digging that out.

jenkins_pipelines/environments/common/pipeline-build-validation.groovy
```diff
@@ -59,6 +59,11 @@ def run(params) {
...
```
maximenoel8: Probably better to avoid same name shadowing.
Proposition:

def cucumberIdleTimeoutMinutes = (params.cucumber_idle_timeout ?: 180).toString().toInteger()

maximenoel8: 180 Minutes idle timeout look excessive in most part. WE should probably analyze a full Jenkins log output to determine those timeout. I also think a per stage timeout make more sense.

maximenoel8 (changes requested):
Looks good. 
my changes requested is related to: the cucumber_idle_timeout  = params.cucumber_idle_timeout.

They leave in different namespaces, so per say not a problem but for readability


Other part not requiring change:
The other will be a small analysis of Jenkins logs to tailor a better timeout. This is a idle timeout, not a full timeout so we could be more strict. 

Anyway good improvement overall, I learned about the activity parameter that I will reuse in an other PR

jenkins_pipelines/environments/common/pipeline-build-validation.groovy
ktsamis: Renamed to `cucumberIdleTimeoutMinutes` in all six pipelines. Kept without `def` in the two build validation files, where it is read from `runCucumberRakeTarget()` and `clientTestingStages()` outside `run()`; the other four use `def`.

ktsamis: Analysed full runs. Worst silence between two console lines, per stage:

| run | length | worst silence | where |
|---|---|---|---|
| manager-4.3-qe-build-validation #43 | 11.6h | 62.4 min | Sync. products and channels |
| manager-5.1-sles-qe-build-validation #9 | 9.0h, SUCCESS | 31.1 min | Sync. products and channels |

Every other build validation stage stayed under 5 minutes, so 180 was indeed excessive almost everywhere, but a flat 60 would have killed 4.3 #43's sync at 62 minutes.

So it is per stage now: 60 everywhere, 180 only for the product sync via `product_sync_idle_timeout`.

One more thing the logs showed: manager-5.2-sles-qe-build-validation #34 ran 67 hours and was never silent for more than 5.5 minutes, retrying `Add Activation Keys` in a loop. An idle timeout cannot catch that one, it needs a wall-clock stage timeout.

jenkins_pipelines/environments/common/pipeline-build-validation-aws.groovy
```diff
@@ -212,15 +226,19 @@ def run(params) {
...
                         // Get minion list from terraform state list command
                         def nodesHandler = getNodesHandler()
-                        res_sync_products = sh(script: "./terracumber-cli ${common_params} --logfile ${resultdirbuild}/testsuite.log --runstep cucumber --cucumber-cmd 'unset ${nodesHandler.envVariableListToDisable.join(' ')}; ${env.exports} cd /root/spacewalk/testsuite; rake cucumber:build_validation_reposync'", returnStatus: true)
+                        withIdleTimeoutOf(productSyncIdleTimeoutMinutes) {
+                            res_sync_products = sh(script: "./terracumber-cli ${common_params} --logfile ${resultdirbuild}/testsuite.log --runstep cucumber --cucumber-cmd 'unset ${nodesHandler.envVariableListToDisable.join(' ')}; ${env.exports} cd /root/spacewalk/testsuite; rake cucumber:build_validation_reposync'", returnStatus: true)
+                        }
                         echo "Custom channels and MU repositories status code: ${res_products}"
                         echo "Custom channels and MU repositories synchronization status code: ${res_sync_products}"
```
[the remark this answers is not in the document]
ktsamis: fixed in https://github.com/SUSE/susemanager-ci/pull/2182, out of scope for this PR

jenkins_pipelines/environments/common/pipeline-rke2.groovy
```diff
@@ -17,6 +17,11 @@ def run(params) {
             def product_commit = null
             def mirror_scope = env.JOB_BASE_NAME.split('-acceptance-tests')[0]
             mirror_scope = mirror_scope.replaceAll("-dev", "")
+            // Inactivity timeout: kills a cucumber run that has stopped producing output entirely.
+            def cucumberIdleTimeoutMinutes = (params.cucumber_idle_timeout ?: 60).toString().toInteger()
+            def withIdleTimeout = { Closure body ->
+                timeout(activity: true, time: cucumberIdleTimeoutMinutes, unit: 'MINUTES') { body() }
+            }
```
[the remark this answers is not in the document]
ktsamis: Not a defect, and not touched by this PR. In a pipeline a variable set through `env.` is also resolvable bare, so `${exports}` and `${env.exports}` are the same value — exactly like `${common_params}` in this file, which is set as `env.common_params` on line 9 and then read bare in every terracumber call that runs today. The mix of both spellings comes from 7402e4ac and is unchanged here.

maximenoel8 (approved):

PR #2189, opened 2026-08-17
Stop deploying the PXE boot minion in the 4.3 test suites

ktsamis:
Stops deploying the PXE boot minion in the 4.3 acceptance test suites (NUE and SLC), so the Cobbler PXE boot and Retail mass import features are skipped instead of failing on every run.

Both features can no longer pass on 4.3. The Manager-4.3 testsuite only maps `sles15sp6o` and `sles15sp7o` to a Kiwi profile, while the 4.3 environments provide a SLES 15 SP4 terminal, and the Cobbler scenario writes the PXE menu entry on the server while the terminal boots from the traditional proxy. They are also the only consumers of the PXE boot minion here: `@skip_if_containerized_server` keeps them out of every other product line, and the Retail PXE boot feature has been skipped since the 4.3 build host was commented out. With no `pxeboot_minion` host, `PXEBOOT_MAC` stays unset and every `@pxeboot_minion` scenario is skipped.

```
Is sles15sp4o a supported image version? (RuntimeError)
./features/support/retail.rb:62:in 'Object#compute_kiwi_profile_name'
```


Bischoff (commented):
I don't like this - it's a bit hiding the dust under the rug :/

Bischoff (changes requested):
Let's drop this, please.

PR #2191, opened 2026-08-20
QE: Add non-installation flags, correct registries and the steps

Pablogoliva:
Add non-installation flags, correct registries and the steps to install uyuni/MLM via testsuite.

### DIFF
--- a/jenkins_pipelines/environments/common/pipeline-rke2.groovy
+++ b/jenkins_pipelines/environments/common/pipeline-rke2.groovy
@@ -72,14 +72,24 @@ def run(params) {
                     // Collect and tag Flaky tests from the GitHub Board
                     def statusCode = sh script: "./terracumber-cli ${common_params} --logfile ${resultdirbuild}/testsuite.log --runstep cucumber --cucumber-cmd 'cd /root/spacewalk/testsuite; ${env.exports} rake utils:collect_and_tag_flaky_tests'", returnStatus: true
                 }
-                stage('Sanity Check') {
+                stage('Install RKE2') {
                     withIdleTimeout {
-                        sh "./terracumber-cli ${common_params} --logfile ${resultdirbuild}/testsuite.log --runstep cucumber --cucumber-cmd 'cd /root/spacewalk/testsuite; ${env.exports} rake cucumber:sanity_check'"
+                        sh "./terracumber-cli ${common_params} --logfile ${resultdirbuild}/testsuite.log --runstep cucumber --cucumber-cmd 'cd /root/spacewalk/testsuite; ${exports} rake cucumber:install_rke2'"
+                    }
+                }
+                stage('Install MLM on RKE2') {
+                    withIdleTimeout {
+                        sh "./terracumber-cli ${common_params} --logfile ${resultdirbuild}/testsuite.log --runstep cucumber --cucumber-cmd 'cd /root/spacewalk/testsuite; ${exports} rake cucumber:install_mlm_on_rke2'"
                     }
                 }
                 stage('Kubernetes sanity checks') {
                     withIdleTimeout {
                         sh "./terracumber-cli ${common_params} --logfile ${resultdirbuild}/testsuite.log --runstep cucumber --cucumber-cmd 'cd /root/spacewalk/testsuite; ${exports} rake cucumber:kubernetes_sanity_checks'"
+                    }
+                }
+                stage('Sanity Check') {
+                    withIdleTimeout {
+                        sh "./terracumber-cli ${common_params} --logfile ${resultdirbuild}/testsuite.log --runstep cucumber --cucumber-cmd 'cd /root/spacewalk/testsuite; ${exports} rake cucumber:sanity_check'"
                     }
                 }
                 stage('Kubernetes tests') {
--- a/terracumber_config/tf_files/MLM-Head-RKE2-NUE.tf
+++ b/terracumber_config/tf_files/MLM-Head-RKE2-NUE.tf
@@ -122,6 +122,7 @@ module "cucumber_testsuite" {
   deploy_hub_api                 = true
   install_kubectl_helm           = false
   kubeconfig_path                = null
+  install_uyuni_via_testsuite    = true
 
   // Cucumber repository configuration for the controller
   git_username = var.GIT_USER
--- a/terracumber_config/tf_files/personal/main_rke2_head.tf
+++ b/terracumber_config/tf_files/personal/main_rke2_head.tf
@@ -70,6 +70,7 @@ module "cucumber_testsuite" {
   deploy_tftp                    = true
   install_kubectl_helm           = false
   kubeconfig_path                = null
+  install_uyuni_via_testsuite    = true
 
   # when changing images, please also keep in mind to adjust the image matrix in the "Used image versions" section of the README.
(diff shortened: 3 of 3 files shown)


terracumber_config/tf_files/personal/main_rke2_head.tf
```diff
@@ -92,7 +93,7 @@ module "cucumber_testsuite" {
       large_deployment     = true
       runtime              = "rke2"
       container_tag        = "latest"
-      container_registry   = "registry.suse.de/suse/containers/suse-multilinuxmanager/5.2/containers/suse/multi-linux-manager/5.2/x86_64/server"
+      container_registry   = "registry.suse.de/suse/containers/suse-multilinuxmanager/5.2/containers/suse/multi-linux-manager/5.2/x86_64"
```
Pablogoliva, line 96: Those are the correct registries. They were wrong.

terracumber_config/tf_files/personal/main_rke2_head.tf
```diff
@@ -105,7 +106,7 @@ module "cucumber_testsuite" {
       main_disk_size       = 200
       runtime              = "rke2"
       container_tag        = "latest"
-      container_registry   = "registry.suse.de/suse/containers/suse-multilinuxmanager/5.2/containers/suse/multi-linux-manager/5.2/x86_64/proxy"
+      container_registry   = "registry.suse.de/suse/containers/suse-multilinuxmanager/5.2/containers/suse/multi-linux-manager/5.2/x86_64"
```
Pablogoliva, line 109: Those are the correct registries. They were wrong.

jenkins_pipelines/environments/common/pipeline-rke2.groovy
```diff
@@ -72,9 +72,15 @@ def run(params) {
                     // Collect and tag Flaky tests from the GitHub Board
                     def statusCode = sh script: "./terracumber-cli ${common_params} --logfile ${resultdirbuild}/testsuite.log --runstep cucumber --cucumber-cmd 'cd /root/spacewalk/testsuite; ${env.exports} rake utils:collect_and_tag_flaky_tests'", returnStatus: true
                 }
+                stage('Install RKE2') {
+                    sh "./terracumber-cli ${common_params} --logfile ${resultdirbuild}/testsuite.log --runstep cucumber --cucumber-cmd 'cd /root/spacewalk/testsuite; ${exports} rake cucumber:install_rke2'"
+                }
+                stage('Install MLM on RKE2') {
+                    sh "./terracumber-cli ${common_params} --logfile ${resultdirbuild}/testsuite.log --runstep cucumber --cucumber-cmd 'cd /root/spacewalk/testsuite; ${exports} rake cucumber:install_mlm_on_rke2'"
+                }
```
ktsamis, line 84: These are the only cucumber stages in this file without `withIdleTimeout`, and they are the two most likely to hang: transactional reboots, `helm --wait`, and the 15 minute readiness waits in `srv_rke2_check_installation.feature`. A hang here eats the whole job timeout instead of failing the stage.

Please wrap both `sh` calls the same way `Sanity Check` does just below.

ktsamis (changes requested):
Registries check out. `MLM-Head-RKE2-NUE.tf` already uses the same `.../x86_64` form with no `/server` or `/proxy` suffix, so the personal file was wrong.

Order: https://github.com/uyuni-project/sumaform/pull/2270 has to merge before this one, otherwise `install_uyuni_via_testsuite` is not a known module variable and terraform fails on the RKE2 job. I have asked for that variable to be renamed there, so both tf files here will need to follow.

One more thing worth knowing: `pipeline-rke2.groovy` is shared with `uyuni-master-dev-acceptance-tests-RKE2`, which runs `Uyuni-Master-RKE2-NUE.tf` on a Tumbleweed server and does not get the flag. The two new stages are harmless there only because the new features are tagged `@transactional_server`. `srv_rke2_check_installation.feature` is not tagged that way though, and the `install_mlm_on_rke2` run set pulls it in, so that job will now run the Kubernetes readiness checks twice.

jenkins_pipelines/environments/common/pipeline-rke2.groovy
Pablogoliva: Done

terracumber_config/tf_files/personal/main_rke2_head.tf
```diff
@@ -70,6 +70,7 @@ module "cucumber_testsuite" {
   deploy_tftp                    = true
   install_kubectl_helm           = false
   kubeconfig_path                = null
+  install_uyuni_via_testsuite    = false
```
ktsamis: Should this be `true`?

The server here is `slmicro62o`, and with the flipped polarity in sumaform #2270 `false` means sumaform installs MLM. But every install state is skipped on SL Micro 6.2, so nothing installs it and the controller never exports `UYUNI_NOT_INSTALLED`. The environment comes up with no server and a testsuite that thinks there is one.

terracumber_config/tf_files/MLM-Head-RKE2-NUE.tf
```diff
@@ -122,6 +122,7 @@ module "cucumber_testsuite" {
   deploy_hub_api                 = true
   install_kubectl_helm           = false
   kubeconfig_path                = null
+  install_uyuni_via_testsuite    = true
```
ktsamis, line 125: Not blocking, just noting that `install_mlm_server = true` ten lines up reads as a contradiction with this. It works only because the sumaform states are skipped on SL Micro 6.2.

ktsamis (commented):
Re-reviewed against the current head. The idle timeouts are in, and the registry change checks out: the helm values `repository:` is a prefix the chart appends image names to (`values_server.yaml:46`, `values_proxy.yaml:8`), so the `/server` and `/proxy` suffixes were wrong. The proxy chart alone pulls five images. Two notes on the flag.

terracumber_config/tf_files/personal/main_rke2_head.tf
ktsamis: Confirmed, the `repository:` value is a prefix the chart appends image names to, so the `/server` suffix was wrong. Resolving.

terracumber_config/tf_files/personal/main_rke2_head.tf
ktsamis: Same here, the proxy chart pulls five images off this prefix. Resolving.

terracumber_config/tf_files/personal/main_rke2_head.tf
ktsamis: @Pablogoliva ping on this one, it is the last blocker from my side.

sumaform#2270 now defaults `install_uyuni_via_testsuite` to `false` meaning sumaform installs, and `MLM-Head-RKE2-NUE.tf:125` sets `true`. This file still says `false` on the same `slmicro62o` server, so it looks like the file the polarity flip was missed on. Should it be `true`?

Pablogoliva: Good catch, chaning it.

Pablogoliva: Done

ktsamis (approved):

PR #2196, opened 2026-08-28
Give the client bootstrap stage its own timeout

ktsamis:
Runs the `Core - Initialize clients` stage with `DEFAULT_TIMEOUT=300` instead of the 250 the rest of the run uses.

Six cucumber workers onboard their clients at the same time, and the server regularly takes longer than 250 seconds to finish `Apply states` under that load. The stage then fails on a timeout that has nothing to do with the change under test, and passes on a retry with no other difference. `pipeline-build-validation.groovy` already overrides the timeout for its bootstrap stage; this applies the same pattern to `pipeline.groovy`, `pipeline-rke2.groovy` and `pipeline-personal.groovy`.

The override only takes full effect together with https://github.com/uyuni-project/uyuni/pull/12521, which stops the wait for an event to appear in the system history from ignoring the caller's timeout.


ktsamis:
Superseded: the timeout now lives in the init_clients feature files instead of the pipeline, in https://github.com/uyuni-project/uyuni/pull/12521.

PR #2208, opened 2026-09-03
Replace prs_checks and the helm_charts_tests with pr-checker

cbosdo:
Gitarro was not maintainable and had design flaws. Rather than sticking with a language that only a handful master, pr-checker switches to python.

This commit uses pr-checker in two pipelines to minimize disruption. If it turns out to be successful, the other pipelines will be replaced too.

### DIFF
--- a/jenkins_pipelines/manager_prs/Jenkinsfile_prs_check
+++ b/jenkins_pipelines/manager_prs/Jenkinsfile_prs_check
@@ -13,14 +13,10 @@ pipeline {
     }
 
     parameters {
-        string(defaultValue: '', description: 'Gitarro PR', name: 'GITARRO_PR_NUMBER')
     }
 
     environment {
       repository = 'SUSE/spacewalk'
-      gitarro_cmd = 'gitarro.ruby2.5'
-      gitarro_local = 'ruby gitarro.rb'
-      check = " -r ${repository} -t placeholder --check --changed_since 172800"
     }
 
     agent { label 'suse-manager-unit-tests' }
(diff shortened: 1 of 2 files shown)


Bischoff (approved):

jenkins_pipelines/manager_prs/Jenkinsfile_prs_helm_charts_tests
```diff
@@ -12,15 +12,12 @@ pipeline {
...
-        string(defaultValue: '', description: 'Gitarro PR', name: 'GITARRO_PR_NUMBER')
         string(defaultValue: '', description: 'SUSE Manager PR', name: 'PR_NUMBER')
         booleanParam(defaultValue: true, description: 'Clean up workspace after a successful execution.', name: 'cleanWorkspace')
     }
 
     environment {
-        gitarro_cmd = 'gitarro.ruby2.5'
-        gitarro_local = 'ruby gitarro.rb'
-        test_script = 'susemanager-utils/testing/automation/helm-charts-tests.sh'
+        test_script = ''
```
ktsamis: `test_script` (and the rest of this `environment` block, `helmcharts_test`) becomes dead code once the `helmcharts_test_cmd` line further down is removed. Also, the new `pr-checker --repo ${repository}` call references `repository`, which isn't defined anywhere in this file. Suggest replacing the whole block with just `repository = 'SUSE/spacewalk'` (matching `Jenkinsfile_prs_check`) - the suggestion below only covers this line since the rest of the block falls outside this diff's visible range, but `helmcharts_test` and the remaining lines should be deleted too. Worth noting while you're at it: pr-checker matches file filters with `fnmatch` (glob syntax, per `pr_checker.py`), so this leftover `-f "containers/.*-helm/.*"` regex-style pattern would be wrong syntax anyway if it were ever revived - the glob form used in the new `pr-checker-config.json` (`containers/*-helm/*`) is the right one.

```suggestion
        repository = 'SUSE/spacewalk'
```

jenkins_pipelines/manager_prs/Jenkinsfile_prs_helm_charts_tests
```diff
@@ -54,14 +43,19 @@ pipeline {
...
+                            " --log-level debug" +
+                            " --repo ${repository}" +
+                            " run" +
+                            " --pr ${params.PR_NUMBER}" +
+                            " --check-name helm_charts_tests" +
+                            " --command susemanager-utils/testing/automation/helm-charts-tests.sh"
+                        )
+                    } else {
+                        echo "Missing PR_NUMBER parameter"
                     }
```
ktsamis, line 52: This unconditionally reads `gitarro_cmd`, which no longer exists (removed from `environment{}` above), so this stage will crash on every run with `MissingPropertyException` before even reaching the `pr-checker` call. Also `SUSE-Manage` is missing a trailing "r" (it was `SUSE-Manager` before this change), and `currentBuild.displayName` was dropped without a replacement so builds lose their PR-number label in the Jenkins UI.

```suggestion
                script {
                    if (params.PR_NUMBER != '') {
                        currentBuild.displayName = "PR: ${params.PR_NUMBER}"
                        sh(
                            "export PRODUCT=SUSE-Manager && pr-checker" +
                            " --log-level debug" +
                            " --repo ${repository}" +
                            " run" +
                            " --pr ${params.PR_NUMBER}" +
                            " --check-name helm_charts_tests" +
                            " --command susemanager-utils/testing/automation/helm-charts-tests.sh"
                        )
                    } else {
                        echo "Missing PR_NUMBER parameter"
                    }
```

jenkins_pipelines/manager_prs/Jenkinsfile_prs_check
```diff
@@ -36,49 +32,40 @@ pipeline {
...
-                            ).trim()
-                        echo PENDING
-                        if (PENDING.contains('TESTREQUIRED=true')) {
-                            PR_NUMBER = sh(script: "grep 'PR_NUMBER:' ${env.WORKSPACE}/.gitarro_vars | cut -d ':' -f2", returnStdout: true).trim()
+                    writeFile(
+                        file: "${env.WORKSPACE}/pr-checker-config.json",
+                        encoding: "UTF-8",
+                        text: '''\
+                          |{
+                          |  "changelog_test" : "",
```
ktsamis: Not a functional bug (pr-checker's `list_prs` iterates this as `for p in patterns`, and a 0-length string iterates zero times same as `[]`), but every other check here maps to a JSON array while this one is a bare string. Suggest `[]` for consistency with the schema the other 9 entries use.

```suggestion
                          |  "changelog_test" : [],
```

jenkins_pipelines/manager_prs/Jenkinsfile_prs_check
```diff
@@ -36,49 +32,40 @@ pipeline {
...
+                          |  "susemanager_unittests" : ["susemanager/"]
+                          |}
+                        '''.stripMargin(),
+                    )
+                    sh(
+                        "pr-checker --log-level debug --repo ${repository} list --config ${env.WORKSPACE}/pr-checker-config.json"
+                    )
+                    checks = readJSON(file: "${env.WORKSPACE}/checks.json")
+                    checks.each { pr, checks ->
+                        checks.each { context ->
```
ktsamis: `checks` (the outer map) and `checks` (the inner per-PR list, the closure param) shadow each other. Works today, but a future edit that needs the outer map inside the inner closure will silently read the wrong value.

```suggestion
                    checks.each { pr, prChecks ->
                        prChecks.each { context ->
```

ktsamis (changes requested):
A few issues from the gitarro to pr-checker migration - see inline comments and suggestions. Checked against the pr-checker source (cbosdo/pr-checker) to confirm the config schema questions.

jenkins_pipelines/manager_prs/Jenkinsfile_prs_helm_charts_tests
cbosdo: That regex is no longer needed too: I forgot the cleanup of that whole block, thanks for the heads up!

jenkins_pipelines/manager_prs/Jenkinsfile_prs_helm_charts_tests
```diff
@@ -38,30 +29,26 @@ pipeline {
...
                 script {
-                    helmcharts_test_cmd = "${gitarro_cmd} ${helmcharts_test}"
-                    if (params.GITARRO_PR_NUMBER != "") {
-                        helmcharts_test_cmd = "${gitarro_local} ${helmcharts_test}"
-                    }
                     if (params.PR_NUMBER != '') {
-                        helmcharts_test_cmd = "${helmcharts_test_cmd} -P ${params.PR_NUMBER}"
-                        currentBuild.displayName = "PR: ${params.PR_NUMBER}"
+                        sh(
+                            "export PRODUCT=SUSE-Manage && pr-checker" +
```
ktsamis: Two things still outstanding from the previous review: `SUSE-Manage` is still missing the trailing "r" (was `SUSE-Manager` before this migration), and `currentBuild.displayName` was never restored, so builds still lose their PR-number label in the Jenkins UI.

```suggestion
                    if (params.PR_NUMBER != '') {
                        currentBuild.displayName = "PR: ${params.PR_NUMBER}"
                        sh(
                            "export PRODUCT=SUSE-Manager && pr-checker" +
```

jenkins_pipelines/manager_prs/Jenkinsfile_prs_check
```diff
@@ -36,49 +32,40 @@ pipeline {
...
+                          |  "susemanager_unittests" : ["susemanager/"]
+                          |}
+                        '''.stripMargin(),
+                    )
+                    sh(
+                        "pr-checker --log-level debug --repo ${repository} list --config ${env.WORKSPACE}/pr-checker-config.json"
+                    )
+                    checks = readJSON(file: "${env.WORKSPACE}/checks.json")
+                    checks.each { pr, contexts ->
+                        contexts.each { context ->
```
ktsamis: This doesn't match what `pr-checker list` actually writes to `checks.json`. `list_prs()` in pr_checker.py produces `{ "<pr>": {"head_sha": "...", "checks_to_run": ["check1", "check2"]} }` - so `contexts` here is a map (`head_sha`/`checks_to_run` keys), not a list of check names. `contexts.each { context -> ... }` with a single closure param iterates `Map.Entry` objects, so `context` becomes e.g. `head_sha=abc123`, and `build(job: "manager-prs-${context}-pipeline")` resolves to a non-existent job name for every PR that needs a check. Since `build()` resolves the job synchronously even with `wait: false`, this throws and fails the whole stage on the first PR that needs any check - the actual check-triggering never happens.

```suggestion
                    checks.each { pr, prData ->
                        prData.checks_to_run.each { context ->
```

ktsamis (changes requested):
Thanks for the cleanup - the earlier gitarro_cmd crash, undefined repository, checks/context shadowing and filter-syntax points are all resolved. Two of the smaller items from before are still outstanding, plus I found a more serious one while re-checking against pr-checker's actual list output (pr_checker.py's list_prs writes {pr: {head_sha, checks_to_run}}, not {pr: [check, ...]}).

ktsamis (approved):
Thanks a lot for this!

PR #2216, opened 2026-09-07
Make "base_arm" a first class citizen

Bischoff:
Make "base_arm" a first class citizen

See also https://github.com/uyuni-project/sumaform/pull/2277 .

DO NOT MERGE. I am not convinced of the merits of this PR. Maxime will tell.

### DIFF
--- a/terracumber_config/tf_files/templates/build-validation-multi-providers.tf
+++ b/terracumber_config/tf_files/templates/build-validation-multi-providers.tf
@@ -25,16 +25,22 @@ provider "libvirt" {
   uri   = "qemu+tcp://${var.BASE_CONFIGURATIONS.base_new_sle.hypervisor}/system"
 }
 
+# Debian/Ubuntu Host
+provider "libvirt" {
+  alias = "host_deblike"
+  uri   = "qemu+tcp://${var.BASE_CONFIGURATIONS.base_deblike.hypervisor}/system"
+}
+
 # Retail/Infrastructure Host
 provider "libvirt" {
   alias = "host_retail"
   uri   = "qemu+tcp://${var.BASE_CONFIGURATIONS.base_retail.hypervisor}/system"
 }
 
-# Debian/Ubuntu Host
+# ARM Host
 provider "libvirt" {
-  alias = "host_deblike"
-  uri   = "qemu+tcp://${var.BASE_CONFIGURATIONS.base_deblike.hypervisor}/system"
+  alias = "host_arm"
+  uri   = "qemu+tcp://${var.BASE_CONFIGURATIONS.base_arm.hypervisor}/system"
 }
 
 # Base Core : Core Infra + Main Testsuite images
@@ -86,31 +92,6 @@ module "base_old_sle" {
   }
 }
 
-# Base RedHat : EL and Liberty
-module "base_rhlike" {
-  providers = { libvirt = libvirt.host_old_sle_rhlike }
-  source    = "./modules/base"
-
-  cc_username       = var.SCC_USER
-  cc_password       = var.SCC_PASSWORD
-  product_version   = var.PRODUCT_VERSION != null ? var.PRODUCT_VERSION : var.ENVIRONMENT_CONFIGURATION.product_version
-  name_prefix       = var.ENVIRONMENT_CONFIGURATION.name_prefix
-  use_avahi         = false
-  domain            = var.PLATFORM_LOCATION_CONFIGURATION[var.LOCATION].domain
-
-  images            = var.BASE_CONFIGURATIONS.base_rhlike.images
-
-  ssh_key_path      = var.CONTROLLER_PUBLIC_SSH_KEY_PATH
-  mirror            = var.PLATFORM_LOCATION_CONFIGURATION[var.LOCATION].mirror
-  use_mirror_images = true
-  testsuite         = true
-
-  provider_settings = {
-    pool        = var.BASE_CONFIGURATIONS.base_rhlike.pool
-    bridge      = var.BASE_CONFIGURATIONS.base_rhlike.bridge
-  }
-}
-
 # Base New SLE : SLES 15 + SLES 16 + SL(E) Micro
 module "base_new_sle" {
   providers = { libvirt = libvirt.host_new_sle }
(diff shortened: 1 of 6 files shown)


Pablogoliva (approved):
LGTM

terracumber_config/tf_files/templates/build-validation-single-provider.tf
```diff
@@ -60,6 +60,7 @@ module "build_validation_module" {
     libvirt.host_rhlike  = libvirt
     libvirt.host_deblike = libvirt
     libvirt.host_retail  = libvirt
+    libvirt.host_arm     = libvirt
```
ktsamis, line 63: `libvirt` is the x86 core hypervisor, so every single-provider env with ARM minions (`mlmhead_nue`, `mlm52_sles_prg`, `mlm51_sles_nue`, `mlm_sandbox_nue`, `suma43_nue`, `suma50_micro_nue`) would create aarch64 guests there. This template also needs its own `host_arm` provider (`try(var.BASE_CONFIGURATIONS.base_arm.hypervisor, "suma-arm.mgr.suse.de")`, since 5 tfvars have no `base_arm`), a `module "base_arm"` and `arm = module.base_arm.configuration`.

```suggestion
    libvirt.host_arm     = libvirt.host_arm
```

terracumber_config/tf_files/templates/build-validation-multi-providers.tf
```diff
@@ -173,16 +205,18 @@ module "base_deblike" {
   use_avahi         = false
   domain            = var.PLATFORM_LOCATION_CONFIGURATION[var.LOCATION].domain
 
-  images            = var.BASE_CONFIGURATIONS.base_deblike.images
+  images            = var.BASE_CONFIGURATIONS.base_arm.images
```
ktsamis, line 208: `suma43_build_validation_slc.tfvars` has no `images` in `base_arm`, so the 4.3 BACKUP plan errors here.

```suggestion
  images            = lookup(var.BASE_CONFIGURATIONS.base_arm, "images", ["opensuse156armo", "opensuse160armo", "raspios13o"])
```

terracumber_config/tf_files/templates/build-validation-multi-providers.tf
```diff
@@ -173,16 +205,18 @@ module "base_deblike" {
...
 
   ssh_key_path      = var.CONTROLLER_PUBLIC_SSH_KEY_PATH
-  mirror            = var.PLATFORM_LOCATION_CONFIGURATION[var.LOCATION].mirror
+  # Pick Prague's mirror even for a test suite running in Salt Lake City
+  mirror            = var.PLATFORM_LOCATION_CONFIGURATION["prg2"].mirror
```
ktsamis: `nue` and `prg2` resolve to the same mirror, so this really means "not the SLC one" — worth saying that instead, and note `domain` right above still follows `var.LOCATION`.

terracumber_config/tf_files/templates/build-validation-multi-providers.tf
```diff
@@ -208,6 +243,7 @@ module "build_validation_module" {
     rhlike  = module.base_rhlike.configuration
     deblike = module.base_deblike.configuration
     retail  = module.base_retail.configuration
+    arm     = module.base_arm.configuration
```
ktsamis, line 245: sumaform#2277 doesn't consume this: `local.base_arm` is unused there and the ARM resources still use the module's own `module "base_arm"`. So the base gets built twice on the same host/pool and collides on the `${name_prefix}${image}` volumes.

ktsamis (changes requested):
Needs uyuni-project/sumaform#2277 merged first — the red `Validate sumaform files` is that (the workflow checks out sumaform `master`). Note that workflow only exercises the single-provider template, so the multi-provider issues below aren't covered by CI.

terracumber_config/tf_files/templates/build-validation-multi-providers.tf
ktsamis: ah but 4.3 is dead, maybe there is no need here

Bischoff: ok good catch, will take care of that, but not with the suggested change - rather in suma43_build_validation_slc.tfvar .

terracumber_config/tf_files/templates/build-validation-multi-providers.tf
Bischoff: "nue" is a leftover that is meant to disappear, we have nothing in Nuremberg anymore, and it really means "Prague". In any case, that will go, but for other reasons.

terracumber_config/tf_files/templates/build-validation-multi-providers.tf
Bischoff: I have doubts about this : if true, why does it work for the others (for example retail) ?

Also, I see it only once:
```
jenkins@b025879e82bb:~/workspace/manager-5.2-sles-qe-build-validation-BACKUP/results/sumaform> tofu state list | grep base.*volumes | grep 160
module.base_core.module.base_backend.libvirt_volume.volumes["opensuse160o"]
module.base_new_sle.module.base_backend.libvirt_volume.volumes["sles160o"]
module.base_retail.module.base_backend.libvirt_volume.volumes["opensuse160o"]
module.build_validation_module.module.base_arm.module.base_backend.libvirt_volume.volumes["opensuse160armo"]
```

terracumber_config/tf_files/templates/build-validation-single-provider.tf
Bischoff: ... but then
    libvirt.host_deblike = libvirt
would have same problem ?

also
  libvirt.host_arm     = libvirt.host_arm
really sounds circular to me (although we have it somewhere else too)

terracumber_config/tf_files/templates/build-validation-multi-providers.tf
Bischoff: Done, there were other files affected

terracumber_config/tf_files/templates/build-validation-multi-providers.tf
Bischoff: Gone anyways.

terracumber_config/tf_files/templates/build-validation-multi-providers.tf
Bischoff: Left unchanged.

Bischoff:
Too complicated, too little benefit

PR #2219, opened 2026-09-07
Use minima mirror in Prague for suma-arm

Bischoff:
Always use minima mirror in Prague for suma-arm hypervisor in Prague,
even though it runs test suites both in Prague and Salt Lake city.

Note: we might have same case for s390. We have no mainframe in SLC...


ktsamis (approved):

Bischoff:
It can't work without adding "mirror" field to the BASE_CONFIGURATIONS map, abandoning the idea.

PR #2237, opened 2026-09-14
Remove Gitarro from uyuni_prs pipelines

cbosdo:
There was a tiny island of remaining Gitarro resisting pipelines.
Gitarro zombie


jenkins_pipelines/uyuni_prs/Jenkinsfile_prs_check
```diff
@@ -36,46 +29,39 @@ pipeline {
...
+                    )
+                    sh(
+                        "pr-checker --log-level debug --repo ${repository} list --config ${env.WORKSPACE}/pr-checker-config.json"
+                    )
+                    checks = readJSON(file: "${env.WORKSPACE}/checks.json")
+                    checks.each { pr, prData ->
+                        prData.checks_to_run.each { context ->
                             build(
-                                job: "uyuni-prs-${context}-pipeline",
+                                job: "manager-prs-${context}-pipeline",
```
ktsamis: Blocker: this is the uyuni dispatcher (`repository = "uyuni-project/uyuni"`), but it now triggers the `manager-prs-*` jobs. Looks like it got copied over from `manager_prs/Jenkinsfile_prs_check`, where `manager-prs-` is correct.

Nothing will error, because those jobs all exist and take a `PR_NUMBER` too, so it'll silently feed uyuni PR numbers into the SUSE/spacewalk pipelines.

```suggestion
                                job: "uyuni-prs-${context}-pipeline",
```

jenkins_pipelines/uyuni_prs/Jenkinsfile_prs_susemanager_unittests
```diff
@@ -8,29 +8,28 @@ properties([
...
         description = "susemanager unit test"
-        filter = "susemanager/"
         git_fs = "${env.WORKSPACE}"
-        // the actual test is the git repo
         test = "susemanager-utils/testing/automation/susemanager-unittest.sh"
-        gitarro_common_params = "-r ${repository} -c ${context} -d ${description} -f ${filter} -t ${test} -g ${git_fs}"
-        gitarro_cmd = 'gitarro.ruby2.5'
-        gitarro_local = 'ruby gitarro.rb'
-        runtest = "${gitarro_common_params} -u ${env.BUILD_URL}"
+        runtest = "pr-checker --repo ${repository} run " +
```
ktsamis: `git_fs` is a Gitarro leftover, nothing reads it any more. Also `run " +` leaves a double space in the command line.

```suggestion
        test = "susemanager-utils/testing/automation/susemanager-unittest.sh"
        runtest = "pr-checker --repo ${repository} run" +
```

jenkins_pipelines/uyuni_prs/Jenkinsfile_prs_schema_migration_test_pgsql
```diff
@@ -8,33 +8,25 @@ properties([
...
-                " -f ${filter} " +
-                " -g ${git_fs} " +
-                " -u \"${env.BUILD_URL}\"" +
-                " -t ${test} "
+        runtest_pg = "pr-checker --repo ${repository} run" +
+                " --pr ${params.PR_NUMBER}" +
+                " --check-name ${context}" +
+                " --description \"${description}\"" +
+                " --build-url ${env.BUILD_URL}" +
+                " --command ${test}"
```
ktsamis: This is the only one of the five without `--git-dir`, so pr-checker clones into a `TemporaryDirectory` outside the workspace. Nothing breaks today since this pipeline collects no JUnit results, but if it's not deliberate it'll bite the first time someone wants artifacts or a post-mortem of the clone.

```suggestion
                " --command ${test}" +
                " --git-dir ${env.WORKSPACE}/clone"
```

ktsamis (commented):
Nice cleanup, the pipelines read much better without the Gitarro branches. One blocker inline plus a couple of nits.

Two things that aren't tied to a line:

**Stuck `pending` statuses have no escape hatch now.** In pr-checker's `evaluate_check_run` the `pending` guard runs *before* the magic-comment and re-run-checkbox checks. So if a job dies between posting `pending` and posting the final status (abort, timeout, worker loss, or the final REST call itself erroring out), the context stays pending forever and neither `- [x] Re-run test "..."` nor `rerun <check> !!!` can clear it. Gitarro's `--changed_since 172800` gave us some self-healing here and nothing replaces it. Probably worth moving the pending guard below the two manual overrides in pr-checker rather than changing anything here.

**git version on the workers.** pr-checker's `run` uses `git clone --revision=<sha>`, which needs git >= 2.49. SLE 15 / Leap 15.6 ship 2.43. Worth double-checking what the `suse-manager-unit-tests` agents actually have before this goes in, since every uyuni check now depends on it.

Also purely cosmetic: `Jenkinsfile_prs_backend_unittests_pgsql` builds the command inline in the `script` block while the other three build it in `environment {}`. Not worth a respin on its own, but the inconsistency invites divergence later.

jenkins_pipelines/uyuni_prs/Jenkinsfile_prs_check
cbosdo: oops! Copy / paste issue!

